PR #278, opened 2023-08-08
improve container credentials retrieval matching container repository names

### LINKED ISSUE #224: Improve container credentials retrieval matching container repository names
Wave retrieves the container credentials to be used, querying Tower credentials for `container-reg` provider and looking for a matching registry name for the given repository e.g. `docker.io` given the repository `docker.io/library/ubuntu`. 

However, this approach limits the possibility of authenticating repositories in the same registry owned by different users or organisations, requiring different credentials. 

This issue aims to extend the credentials retrieval mechanism so that the entire container repository name is used to discover the matching credentials to be used, not just the registry name. 

The implementation of this feature also requires that Tower allows, optionally, storing the full container name,

munishchouhan:
1. wave should support repository credential validation
2. Wave should choose the best matching credentials from the tower for a give registry/repository name


marcodelapierre:
@pditommaso  ready to review!

marcodelapierre:
@pditommaso pinging you on this one - can you provide an output on any dependency that is holding this PR? thank you :-)

PR #340, opened 2023-11-28
added CachedThreadPool in TimedInputStream

munishchouhan:
This PR will add CachedThreadPool in TimedInputStream to replace ForkJoinPool.commonPool()


src/main/groovy/io/seqera/wave/util/TimedInputStream.groovy
```diff
@@ -40,18 +42,21 @@ class TimedInputStream extends FilterInputStream {
 
     private final RoutePath route
 
+    private final ExecutorService executor
```
pditommaso, line 45: This can be overkill because it creates a new thread pool every time a new instance is created. This field should be `static` to use only one thread pool instance. 

Also, not sure anymore if the CachedThreadPool is the best choice here. Under heavy load, since the CachedThreadPool is unbound, it can result in the creation of hundreds of threads. 

It my be better to use instead a fixed thread tool with up to 32 threads (default)

pditommaso:
I'm closing this because i've a bit messed with the test branch. Simpler to recreate it

PR #342, opened 2023-11-28
Improve thread pools

pditommaso:
This PR  improves the use of thread pools in the following manner: 

- Use a fixed thread pool for the `TimedInputStream` with 32 threads. The rationale is to use a larger pool compared to the default common thread pool that only uses 8 threads (one for avail CPUs).
- Use a cached thread pool both for the `ContainerBuildServiceImpl` and `ContainerScanServiceImpl` because here latency is not critical, and a few requests are expected.


pditommaso:
Closing in favour of 1cd2dc90

PR #344, opened 2023-11-30
Update build.yml: softer conditional to skip build CI

marcodelapierre:
I think this [commit](https://github.com/seqeralabs/wave/pull/71/commits/bdecfc3793fa98836d1215802fcf48eaaa309a7d) has to be reverted,  because it is causing CI to be skipped, in cases where they should be executed.

Not having the current conditional is in line with nextflow, tower, fusion repos.  

Leaving the condition on `[skip ci]` commits, as in wave-cli, can be good to skip CIs in specific cases.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,7 +24,7 @@ on:
 jobs:
   build:
     name: Build Wave
-    if: "github.event == 'push' || github.repository != github.event.pull_request.head.repo.full_name"
+    if: "!contains(github.event.head_commit.message, '[ci skip]')"
     runs-on: ubuntu-latest
     timeout-minutes: 15
     strategy:


marcodelapierre:
One thing I don't understand is why in this PR of yours there are 3 checks, 2 of which identical (build due to push and build due to pull_request): https://github.com/seqeralabs/wave/pull/342/checks

Whereas in this current PR the push one is not present.
On the other hand there are 2 instances of pull_request, apparently with two distinct IDs .. if it has run twice, it means twice the compute, which is not good I guess.

Why is this the case? 

I don't understand.

@pditommaso how do you perform these actions? local edits, commits, and then push to github? I have done the current one with live editing here in the github repo, because it was small edits.  If these two scenarios are distinct to Github, there should be a way to discern them in the GH workflow, but so far I haven't found it. I will dig again.

PR #358, opened 2023-12-19
Proxy client v4

pditommaso:
Another tentative to implement the binary fetch via the streaming client


pditommaso:
This commit implements the use of ReactorStreamingHttpClient 
to stream the layer blob binaries, in place of the Java SDK HTTP client
which shows several memory issues for large payload. 

Moreover, it uses the setting `wave.httpclient.streamThreshold` (default 64k)
to determine when read the layer blob directly into a `byte[]` buffer or 
use a separate HTTP request and stream the content for large blobs

Issue #338, opened 2023-11-24
Implement Pull-through cache mode


marcodelapierre:
Posting original thoughts by @pditommaso  on this topic.

- In 70-80% of requests, Wave only redirects to the bucket, and then the Docker client performs a direct layer pull from object storage
- However for some containers this path is not taken, and Wave has to "serve" the whole binary payload, which causes issues
- Ideally we should create a cache and make use of it at download time
- https://docs.docker.com/docker-hub/mirror/
- https://github.com/seqeralabs/wave/blob/d29f217066c17834112e65e6c6bf8ebd77763c28/src/main/groovy/io/seqera/wave/controller/RegistryProxyController.groovy#L243-L257

pditommaso:
Solved by #359

PR #359, opened 2023-12-19
Container blob cache

pditommaso:
This PR implements a caching mechanism for layer blobs to achieve to main goals: 
- deduplicate the downloads of the same binary layers 
- offload the download of blob binaries to the S3 bucket. 

The overall logic is very similar to the one implemented by [ContainerBuildServiceImpl](https://github.com/seqeralabs/wave/blob/049a20763b875a6bd6b99bf456a3037a5b329c26/src/main/groovy/io/seqera/wave/service/builder/ContainerBuildServiceImpl.groovy#L67-L67), implementing the download and storing to S3 instead of building the image 

Solves https://github.com/seqeralabs/wave/issues/338


pditommaso:
Ready for testing. The blob caching is enabled by settings in the application config 

```
wave:
  blobCache:
    enabled: true
    baseUrl: "https://cache-bucket-name-hostname"
    storage:
      bucket: "s3://<cache-bucket-name>"
```

When this feature is enabled, Wave caches the container blob binaries in the `cache-bucket-name` bucket specified in the above setting and returns an HTTP 307 redirect response with the URL of the cache blob content. 

The upload to the object storage is made via [s5cmd](https://github.com/seqeralabs/wave/blob/4933e158d3df72f83f5d0f386b36674593c712c5/src/main/resources/application.yml#L74-L74) run in a (pod) container. This avoid overloading the Wave backend with memory-intensive binary transfer operations. 

The use of a CDN network such as CloudFront or Cloudflare can be used to further optimise the caching and download of the container binaries

pditommaso:
To complete this feature, we may need to generate [signed URL](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/examples-s3-presign.html) for the objects stored in the S3 bucket so that they will be accessible only for a specified period of time. 

Likely the logic should be implemented [here](https://github.com/seqeralabs/wave/blob/2e3e847b9750ce7563190a09183d183387065096/src/main/groovy/io/seqera/wave/service/blob/impl/BlobCacheServiceImpl.groovy#L211-L213). 

@munishchouhan want to give a try to this?

munishchouhan:
> To complete this feature, we may need to generate [signed URL](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/examples-s3-presign.html) for the objects stored in the S3 bucket so that they will be accessible only for a specified period of time.
> 
> Likely the logic should be implemented [here](https://github.com/seqeralabs/wave/blob/2e3e847b9750ce7563190a09183d183387065096/src/main/groovy/io/seqera/wave/service/blob/impl/BlobCacheServiceImpl.groovy#L211-L213).
> 
> @munishchouhan want to give a try to this? (please use a separate PR)

sure

marcodelapierre:
Wooo it's merged, well done folks!! 🚀

PR #361, opened 2023-12-26
Added error message when spack object is null

munishchouhan:
This PR will add a message when the spack recipe object is null and throws NPE


marcodelapierre:
Couldn't this happen with Conda recipes as well?

pditommaso:
Well, everything is possible.

pditommaso:
Add the corresponding test please

munishchouhan:
@pditommaso should we move `guessSpackRecipeName(String spackFileContent)` from RegHelper to SpackHelper class?
I think it belongs to the latter class

pditommaso:
Let's keep it here and for consistency with the issue

munishchouhan:
It returns null for every exception
https://github.com/seqeralabs/wave/blob/757ac62a7b6155516a9394f35bdc842951ab69a9/src/main/groovy/io/seqera/wave/util/RegHelper.groovy#L249-L251
Do we want to throw exception when spack and spack.specs is not present or still return null?

src/main/groovy/io/seqera/wave/util/RegHelper.groovy
```diff
@@ -221,6 +222,14 @@ class RegHelper {
         try {
             final yaml = new Yaml().load(spackFileContent) as Map
             final spack = yaml.spack as Map
+
+            if( !spack ){
+                throw new BadRequestException("spack section is not present in your spack file, please add spack section and try again")
```
marcodelapierre, line 227: Rewording suggestion:
```
throw new BadRequestException("Section \"spack:\" not found in the Spack environment file, please check it and try again")
```

src/main/groovy/io/seqera/wave/util/RegHelper.groovy
```diff
@@ -221,6 +222,14 @@ class RegHelper {
...
+            }
+            if( !spack.specs ){
+                throw new BadRequestException("spack.specs section is not present in your spack file, please add spack.specs section and try again")
```
marcodelapierre, line 230: Rewording suggestion:
```
throw new BadRequestException("Section \"spack:specs:\" not found in the Spack environment file, please check it and try again")
```

src/test/groovy/io/seqera/wave/util/RegHelperTest.groovy
```diff
@@ -235,6 +236,25 @@ class RegHelperTest extends Specification {
         RegHelper.guessSpackRecipeName(SPACK) == 'bwa-0.7.15_salmon-1.1.1_nano-1.0'
     }
 
+    def 'should return null whenspack section is not present in spack yaml file' () {
```
marcodelapierre, line 239: typo `whenspack` should be `when spack`

marcodelapierre (changes requested):
Some minor edit suggestions for the error messages, and a minimal typo

marcodelapierre:
> Do we want to throw exception when spack and spack.specs is not present or still return null?

Returning null is consistent with other Spack and Conda issues in the yaml file. I think it is consistent with the scope of this PR. 

If we want to add methods to do a minimal validation of the contents of Spack/Conda yamls, I think it should be in a separate PR.

src/main/groovy/io/seqera/wave/util/RegHelper.groovy
munishchouhan: done

src/main/groovy/io/seqera/wave/util/RegHelper.groovy
munishchouhan: done

src/test/groovy/io/seqera/wave/util/RegHelperTest.groovy
munishchouhan: done

src/main/groovy/io/seqera/wave/util/RegHelper.groovy
```diff
@@ -221,6 +222,14 @@ class RegHelper {
         try {
             final yaml = new Yaml().load(spackFileContent) as Map
             final spack = yaml.spack as Map
+
+            if( !spack ){
+                throw new BadRequestException("Section \"spack:\" not found in the Spack environment file, please check it and try again")
+            }
+            if( !spack.specs ){
+                throw new BadRequestException("Section \"spack:specs:\" not found in the Spack environment file, please check it and try again")
+            }
```
pditommaso, line 231: This is a kind of useless, because the exception is thrown in the `catch` statement and the method just return null. Therefore the error message is never returned the user making the request

PR #362, opened 2024-01-03
Added Check for build success when BuildResult is present in cache

munishchouhan:
This PR adds a check whether the build result present is the cache succeed 
If not, then launch build request


munishchouhan:
Add Unit tests

munishchouhan:
Likely completed by https://github.com/seqeralabs/wave/pull/421

PR #366, opened 2024-01-08
Added tolerations for k8s pod

### LINKED ISSUE #291: [feature] Add option to set toleration for tainted nodes
There are some node groups in the prod cluster to which wave can start pods to build its images, but currently any pod may end up running on those nodes.
One problem with this is that pods may be scheduled on an ARM node and crash due to architecture incompatibility.

We'd prefer to taint the nodes so only pods running with an explicit toleration would be scheduled on those pods.
@pditommaso mentioned that wave may not have this capability for now, so I'm opening a feature request for this.

munishchouhan:
1. Added toleration configuration variables
2. Added set toleration method in k8s service
3. Added isARM64 an isAMD64 method in ContainerPlatform


src/main/groovy/io/seqera/wave/service/k8s/K8sServiceImpl.groovy
```diff
@@ -94,6 +97,26 @@ class K8sServiceImpl implements K8sService {
     @Nullable
     private String requestsMemory
 
+    @Value('${wave.build.k8s.toleration.arm64.enabled:false}')
```
bebosudo, line 100: Note that tolerations aren't tied to arm64 only, so hardcoding it as under the arm64 section is restrictive.

Moreover, customers may need to add multiple different tolerations to the wave build pods, so we should define a way to add multiple tolerations.

Please read the related [taint and toleration docs](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/).

src/main/groovy/io/seqera/wave/service/k8s/K8sServiceImpl.groovy
```diff
@@ -94,6 +97,26 @@ class K8sServiceImpl implements K8sService {
...
+
+    @Value('${wave.build.k8s.toleration.arm64.value}')
+    @Nullable
+    private String tolerationARM64Value
+
+    @Value('${wave.build.k8s.toleration.arm64.operator:`Equal`}')
+    @Nullable
+    private String tolerationARM64Operator
+
+    @Value('${wave.build.k8s.toleration.arm64.effect:`NoSchedule`}')
```
bebosudo, line 116:
```suggestion
    @Value('${wave.build.k8s.toleration.arm64.effect}')
```

The effect can't have a default, since the docs say that an empty effect parameter is allowed:
> An empty effect matches all effects with key key1.

bebosudo (commented):
I can't test this code, but in general, I don't think we should hardcode tolerations around ARM64.

I think what we should do is to take whatever (if any) tolerations (because there could be more than just one) the user wants to set on the wave build pods and pass them on when submitting a pod creation request to the k8s control plane, without adding any logic. Let me know if this makes sense

src/main/groovy/io/seqera/wave/service/k8s/K8sServiceImpl.groovy
munishchouhan: Empty is valid for k8s, but we need to think about our use case and what can be a default scenario.
In this case, we want to schedule arm pods in arm nodes.

src/main/groovy/io/seqera/wave/service/k8s/K8sServiceImpl.groovy
bebosudo: Ok, but we're hardcoding a single specific toleration use case, but customers may need to install wave in other environments and may need other (even multiple) tolerations, not just ARM vs AMD nodes.
Think about customer environments where they have GPU nodes, or newer generation CPUs, and they may want to only allow wave build pods in a specific location.
Let me know if this helps.

src/main/groovy/io/seqera/wave/service/k8s/K8sServiceImpl.groovy
bebosudo: > but we need to think about our use case and what can be a default scenario

Sorry, I think we need to think what customers may want to use, instead of hardcoding our deployment usecase into the wave config.

src/main/groovy/io/seqera/wave/service/k8s/K8sServiceImpl.groovy
munishchouhan: You are correct in that part, but in this case, I am checking whether the platform you asked for is AMR and then applying this toleration. 

https://github.com/seqeralabs/wave/blob/d9c1bfd939c6abbb7622e02fb5dce70aa94a948d/src/main/groovy/io/seqera/wave/service/k8s/K8sServiceImpl.groovy#L406-L409

We can provide another set of properties for a general tolerations, which will be applied to all build pods

bebosudo: Then why hardcoding our special ARM usecase? Let's build a generic toleration system and we'll use that in our use case

munishchouhan: because we only want ARM pods to run on nodes with arm taint and not AMD 
or am I missing something?

bebosudo: That's correct.
My point is: why making a special usecase to add tolerations for ARM only and then a another generic system to add other tolerations? That's why I was proposing to create the generic tolerations system only, and get rid of the special ARM usecase.

bebosudo: In a quick call we were able to share our points: we need to have tolerations for each architecture that wave supports (amd64 and arm64), and users can pass multiple tolerations, so the multiple string parameters will need to change to become a single `List<Toleration>` object for each arch.

src/main/resources/application-dev-k8s.yml
```diff
@@ -16,4 +16,18 @@ wave:
...
+        enabled: true
+        arm64:
+          - key : 'arch'
+            value : 'arm64'
+            operator : 'Equal'
+            effect : 'NoSchedule'
+          - key : 'arch'
+            value : 'aarch64'
+            operator : 'Equal'
+            effect : 'NoSchedule'
```
bebosudo, line 32: Wait is this file for? Are these default tolerations?

bebosudo (approved):
Looks good to me, did you test this on k8s? Let me know if you need help testing it in our dev environment.

bebosudo:
CC @marcodelapierre

src/main/resources/application-dev-k8s.yml
munishchouhan: this config is for dev purpose and to show how it will look in the config file

marcodelapierre:
> Looks good to me, did you test this on k8s? Let me know if you need help testing it in our dev environment.

@munishchouhan have you tested? if not, can you discuss with @bebosudo a suitable setup for testing? thank you both

munishchouhan:
Checked the set toleration in local machine
```
% kubectl describe pod  build-fe286c2a9222f54a-91ceb --namespace wave-local
Name:             build-fe286c2a9222f54a-91ceb
Namespace:        wave-local
Priority:         0
Service Account:  default
Node:             docker-desktop/192.168.65.3
Start Time:       Fri, 19 Jan 2024 15:54:31 +0100
Labels:           <none>
Annotations:      <none>
Status:           Succeeded
IP:               10.1.0.94
IPs:
  IP:  10.1.0.94
Containers:
  build-fe286c2a9222f54a-91ceb:
    Container ID:  docker://2664759eea32a0054c0a696a3d210680d2a60c96282e0bb021738bb436a2e28c
    Image:         gcr.io/kaniko-project/executor:v1.19.2
    Image ID:      docker-pullable://gcr.io/kaniko-project/executor@sha256:f913ab076f92f1bdca336ab8514fea6e76f0311e52459cce5ec090c120885c8b
    Port:          <none>
    Host Port:     <none>
    Args:
      --dockerfile
      /Users/munish.chouhan/.wave/build-workspace/fe286c2a9222f54a/Containerfile
      --context
      /Users/munish.chouhan/.wave/build-workspace/fe286c2a9222f54a/context
      --destination
      hrma017/dev:fe286c2a9222f54a
      --cache=true
      --custom-platform
      linux/amd64
      --cache-repo
      hrma017/dev
    State:          Terminated
      Reason:       Completed
      Exit Code:    0
      Started:      Fri, 19 Jan 2024 15:54:31 +0100
      Finished:     Fri, 19 Jan 2024 15:54:53 +0100
    Ready:          False
    Restart Count:  0
    Requests:
      memory:     2400Mi
    Environment:  <none>
    Mounts:
      /Users/munish.chouhan/.wave/build-workspace/fe286c2a9222f54a from build-data (ro,path="fe286c2a9222f54a")
      /kaniko/.docker/config.json from build-data (ro,path="fe286c2a9222f54a/config.json")
      /var/run/secrets/kubernetes.io/serviceaccount from kube-api-access-pgznl (ro)
Conditions:
  Type              Status
  Initialized       True
  Ready             False
  ContainersReady   False
  PodScheduled      True
Volumes:
  build-data:
    Type:          HostPath (bare host directory volume)
    Path:          /Users/munish.chouhan/.wave/build-workspace
    HostPathType:
  kube-api-access-pgznl:
    Type:                    Projected (a volume that contains injected data from multiple sources)
    TokenExpirationSeconds:  3607
    ConfigMapName:           kube-root-ca.crt
    ConfigMapOptional:       <nil>
    DownwardAPI:             true
QoS Class:                   Burstable
Node-Selectors:              <none>
**Tolerations:                 arch=amd64:NoSchedule
                             node.kubernetes.io/not-ready:NoExecute op=Exists for 300s
                             node.kubernetes.io/unreachable:NoExecute op=Exists for 300s**
```

PR #369, opened 2024-01-10
Spack templates: flipped order of mirrors

marcodelapierre:
I believe the Seqera binary mirror should take priority over the Spack team one in our builds; this PR actions this.

@bebosudo  FYI

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,7 +24,6 @@ on:
 jobs:
   build:
     name: Build Wave
-    if: "github.event == 'push' || github.repository != github.event.pull_request.head.repo.full_name"
     runs-on: ubuntu-latest
     timeout-minutes: 15
     strategy:
--- a/src/main/resources/io/seqera/wave/spack/spack-builder-dockerfile.txt
+++ b/src/main/resources/io/seqera/wave/spack/spack-builder-dockerfile.txt
@@ -30,8 +30,8 @@ RUN mkdir -p /opt/spack-env \
 # Install packages, clean afterward, finally strip binaries
 RUN cd /opt/spack-env \
 && fingerprint="$(spack gpg trust {{spack_key_file}} 2>&1 | tee /dev/stderr | sed -nr "s/^gpg: key ([0-9A-F]{16}): secret key imported$/\1/p")" \
-&& spack mirror add seqera-spack {{spack_cache_bucket}} \
 && spack mirror add binary_mirror  https://binaries.spack.io/releases/v0.20 \
+&& spack mirror add seqera_spack {{spack_cache_bucket}} \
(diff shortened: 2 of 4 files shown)


marcodelapierre:
CI is failing, first error is:

```
06:31:07.226 [Test worker] WARN  i.s.w.auth.RegistryAuthServiceImpl - Container registry 'http://localhost:32769/v2/' login FAILED: 401 - response: {"errors":[{"code":"UNAUTHORIZED","message":"authentication required","detail":null}]}
```

@bebosudo  can you help me undertsand?  
Do I need special tokens in my Github account to trigger this type of CIs?

.github/workflows/build.yml
```diff
@@ -24,7 +24,6 @@ on:
 jobs:
   build:
     name: Build Wave
-    if: "github.event == 'push' || github.repository != github.event.pull_request.head.repo.full_name"
```
bebosudo, line 27: Why are we removing this line in this PR?

bebosudo (commented):
I don't think this makes any difference.

Inside the wave build process, Spack will look for packages in its cache to use as dependencies, and these dependencies must be a complete match down to the hash numbers in order to be picked up by spack and reused.
So if ever there will be a package match between the spack mirror and our mirror, the package files must be exactly the same, so I believe swapping the order doesn't "increase the priority" when wave builds packages.

bebosudo:
No idea about the CI error, maybe @munishchouhan can help.

.github/workflows/build.yml
marcodelapierre: Come from merging this other PR I did: https://github.com/seqeralabs/wave/pull/370

Basically, when working on a PR in this very same repo (like this one), none of the conditions is true, and hence the CI is skipped. 

(note: my workflow is commits in my local computer, followed by push to this repo, with PR open)

marcodelapierre: I admit I am getting some head aches in completely understanding why CI workflows in our repos sometimes run and sometimes don't ... happy to take any advice! :-)

marcodelapierre: please review https://github.com/seqeralabs/wave/pull/370 first :-)

marcodelapierre:
> I don't think this makes any difference.
> 
> Inside the wave build process, Spack will look for packages in its cache to use as dependencies, and these dependencies must be a complete match down to the hash numbers in order to be picked up by spack and reused. So if ever there will be a package match between the spack mirror and our mirror, the package files must be exactly the same, so I believe swapping the order doesn't "increase the priority" when wave builds packages.

You are totally right @bebosudo , very good point. In this respect, the binary mirror is different from the registry lookup for recipes, where there might exist duplicates of the same recipe with different instructions (here priority is important).

Will close .

PR #370, opened 2024-01-10
build CI: removing conditional within job definition

marcodelapierre:
[added]  This PR removes the following conditional for the `build` CI workflow:
```
if: "github.event == 'push' || github.repository != github.event.pull_request.head.repo.full_name"
```

Before this PR, when working on a PR within this repo (not a fork), the CI workflow was skipped, which is undesirable.

(@munishchouhan rationale added, thank you)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,7 +24,6 @@ on:
 jobs:
   build:
     name: Build Wave
-    if: "github.event == 'push' || github.repository != github.event.pull_request.head.repo.full_name"
     runs-on: ubuntu-latest
     timeout-minutes: 15
     strategy:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,6 +24,7 @@ on:
 jobs:
   build:
     name: Build Wave
+    if: "github.event == 'push' || github.repository != github.event.pull_request.head.repo.full_name"
     runs-on: ubuntu-latest
     timeout-minutes: 15
     strategy:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,7 +24,7 @@ on:
 jobs:
   build:
     name: Build Wave
-    if: "github.event == 'push' || github.repository != github.event.pull_request.head.repo.full_name"
+#    if: "github.event == 'push' || github.repository != github.event.pull_request.head.repo.full_name"
     runs-on: ubuntu-latest
     timeout-minutes: 15
     strategy:
@@ -33,6 +33,11 @@ jobs:
         java_version: [19]
 
     steps:
+      - name: Debug GH context
+        run: echo "$GITHUB_CONTEXT"
+        env:
+          GITHUB_CONTEXT: ${{ toJson(github) }}
+
       - name: Environment
         run: env | sort
 
--- a/src/main/groovy/io/seqera/wave/proxy/LoginResponse.groovy
+++ b/src/main/groovy/io/seqera/wave/proxy/LoginResponse.groovy
@@ -38,3 +38,4 @@ class LoginResponse {
 
 
 // TEST MARCO
+// TEST 22
--- a/src/main/groovy/io/seqera/wave/proxy/LoginResponse.groovy
+++ b/src/main/groovy/io/seqera/wave/proxy/LoginResponse.groovy
@@ -39,3 +39,4 @@ class LoginResponse {
 
 // TEST MARCO
 // TEST 22
+// test alberto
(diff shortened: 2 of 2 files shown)


munishchouhan:
@marcodelapierre is it still needed?

marcodelapierre:
I did this PR and then also merged it in the other 3x concurrent ones because I needed it. 

I did as separate for better visibility.

marcodelapierre:
@bebosudo  apologies I did not mention early, please review this one first, as it is also merged on the other 3x. 

I kept it separate for better present and future visibility in the repo history.

munishchouhan:
@marcodelapierre can you please add a discription about why this change us required?

munishchouhan:
@marcodelapierre Sorry, but I don't understand the purpose
This CI works fine even if you are working on this repository and not in fork
It get skipped mainly when the change in text files, where there is no executable code

Can you please share the push, where you faced any issue?

marcodelapierre:
Thanks for the comment @munishchouhan  .
You are right, I did some study yesterday, and have more insights.

Apparently I am "the only one" having issues with skipped push CI workflows on the Wave repo. I have compared PRs of mine with PRs from you and Paolo, and can tell the difference.

For instance:
- mine, https://github.com/seqeralabs/wave/pull/371, see I have 2 checks only in the checks tab (no push)
- yours, https://github.com/seqeralabs/wave/pull/372, see you have 3 checks

I did some studying, and found that it might relate to GITHUB_TOKENs. The Wave `build.yml` indeed uses them, crucially for the `Release` step ([from line 75](https://github.com/seqeralabs/wave/blob/291-feature-add-option-to-set-toleration-for-tainted-nodes/.github/workflows/build.yml#L75-L86)), which indeed pushes stuff.  Now compare this fact with this quote from https://docs.github.com/en/actions/security-guides/automatic-token-authentication :

```
When you use the repository's GITHUB_TOKEN to perform tasks, events triggered by the GITHUB_TOKEN, with the exception of workflow_dispatch and repository_dispatch, will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs. For example, if a workflow run pushes code using the repository's GITHUB_TOKEN, a new workflow will not run even when the repository contains a workflow configured to run when push events occur.
```

Now, what I don't get is why the behaviour is different for me vs you and Paolo.  Note I have limited previous experience with Github workflows.

Do you have any comments/advice to share on this topic? Thank you.

munishchouhan:
Here are the steps:
1. Setup git with GitHub access, follow this article
https://diptendud.medium.com/github-setup-on-mac-os-8aec9e46d15c
2. checkout the branch either using git checkout "branch_name" or through GUI in IntelliJ
3. for the push you can use git add <file>, git commit -m "<message>" and git push
or you can use IntelliJ GUI's commit tab to push your changes 
Screenshot 2024-01-12 at 10 52 33

marcodelapierre:
Thanks @munishchouhan . Here is the weird thing, what you describe is indeed pretty much how I do it.

- I do have a local `git` installation, which I use equivalently from either terminal (most of the time) or Visual Studio Code.
- I have setup SSH keys (which enable `git clone git@github.com:<org>/<repo>`), and GPG keys as well for signing of commits
- My typical workflow is
  - create a new local branch
  - work and commit locally
  - push to remote repo, either Seqera original or my fork
  - open PR on that repo through Github website
  - as needed, more local commits followed by pushes to remote

As regards GH workflows, the result is what you see in my example above - push workflows do not appear in the list of checks.

marcodelapierre:
@bebosudo if you have time, can you read the last 3-4 comments, detailing the CI issue I am having, and see if you have any idea to share along with Munish?  Thank you

bebosudo:
> yours, https://github.com/seqeralabs/wave/pull/372, see you have 3 checks

I see two checks on Munish' PR too :shrug:

image

Could you share links to GHA runs where the behavior was different between you and Munish?

marcodelapierre:
Mine   Screenshot 2024-01-16 at 6 31 28 pm

marcodelapierre:
Munish:  Screenshot 2024-01-16 at 6 31 53 pm

marcodelapierre:
Let me also share the GHA runs

marcodelapierre:
GHA Runs
* Mine: https://github.com/seqeralabs/wave/actions?query=branch%3Aupdate%2Fspack-yaml-package-configs
* Munish: https://github.com/seqeralabs/wave/actions?query=branch%3AChange-public-to-public.repo-config

bebosudo:
After 1 hour of debugging we found out that the problem is that we're using  `on.push.branches: *` instead of `on.push.branches: **` in our GHA:
https://stackoverflow.com/a/64635612/5810105

marcodelapierre:
Single vs double asterisk documented on this GH doc page: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

Kudos and thank you @bebosudo  for the patient debugging!

@pditommaso @jordeu going to tweak tue CIs for Wave and Fusion shortly.

PR #368, opened 2024-01-10
spack templates BIS: all configs available in spack.yaml and its dir

marcodelapierre:
Hi @pditommaso, this PR updates the Spack templates for Docker/Singularity, so that all package build configurations are visible in `spack.yaml` and its sister yaml, `compilers.yaml`, in the `/opt/spack` directory. 

This change does not impact the way packages are compiled, but in my opinion is beneficial to users as it improved transparency of how the packages are built, hence improving provenance of the build from the software perspective.

@bebosudo FYI

PR #371, opened 2024-01-10
Update/spack yaml package configs

marcodelapierre:
This PR adds extra configuration in the Spack templates, providing good defaults for a number of common dependency packages.

These lines were taken from two key Spack stacks within the Spack repo, i.e.
* [e4s](https://github.com/spack/spack/blob/develop/share/spack/gitlab/cloud_pipelines/stacks/e4s/spack.yaml)
* [aws-isc](https://github.com/spack/spack/blob/develop/share/spack/gitlab/cloud_pipelines/stacks/aws-isc/spack.yaml)

Good side: I believe these defaults will improve overall stability of package builds at scale, as suggested by their adoption by E4S and AWS teams within the Spack repo.

Bad side: the Spack yaml gets more verbose.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,10 +2,10 @@ name: Wave CI
 on:
   push:
     branches:
-      - '*'
+      - '**'
       - '!refs/tags/.*'
     tags-ignore:
-      - '*'
+      - '**'
     paths-ignore:
       - 'docs/**'
       - mkdocs.yml
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,7 +24,6 @@ on:
 jobs:
   build:
     name: Build Wave
-    if: "github.event == 'push' || github.repository != github.event.pull_request.head.repo.full_name"
     runs-on: ubuntu-latest
     timeout-minutes: 15
     strategy:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,6 +24,7 @@ on:
 jobs:
   build:
     name: Build Wave
+    if: "github.event == 'push' || github.repository != github.event.pull_request.head.repo.full_name"
     runs-on: ubuntu-latest
     timeout-minutes: 15
     strategy:
--- a/src/main/resources/io/seqera/wave/spack/spack-builder-singularityfile.txt
+++ b/src/main/resources/io/seqera/wave/spack/spack-builder-singularityfile.txt
@@ -32,7 +32,21 @@ EOF
     spack config add concretizer:unify:true
     spack config add concretizer:reuse:true
     spack config add packages:all:target:[{{spack_arch}}]
+    spack -e . config add packages:all:providers:blas:[openblas]
+    spack -e . config add packages:all:providers:mkl:[intel-oneapi-mkl]
+    spack -e . config add packages:binutils:variants:"+ld +gold +headers +libiberty ~nls"
+    spack -e . config add packages:elfutils:variants:"+bzip2 ~nls +xz"
+    spack -e . config add packages:hdf5:variants:"+fortran +hl +shared"
+    spack -e . config add packages:libunwind:variants:"+pic +xz"
+    spack -e . config add packages:mesa:variants:~llvm
+    spack -e . config add packages:mesa18:variants:~llvm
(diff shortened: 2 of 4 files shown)


marcodelapierre:
CI Build fails to be investigated yet (will tackle fails in https://github.com/seqeralabs/wave/pull/369 first)

.github/workflows/build.yml
```diff
@@ -24,7 +24,6 @@ on:
 jobs:
   build:
     name: Build Wave
-    if: "github.event == 'push' || github.repository != github.event.pull_request.head.repo.full_name"
```
bebosudo, line 27: Why are we removing this line here too?

bebosudo (approved):
LGTM, but is there a way we could avoid hardcoding these in wave, to help keep them in sync with the configs in the stack files?

FTR, here's where the E4S are defining these options in their spack stack file:
https://github.com/spack/spack/blob/f552dbd19918b6c928dd482e1de92d920902f1f3/share/spack/gitlab/cloud_pipelines/stacks/e4s/spack.yaml#L8-L60

.github/workflows/build.yml
marcodelapierre: Come from merging https://github.com/seqeralabs/wave/pull/370, as I needed to run CI tests appropriately within this PR

marcodelapierre: please review https://github.com/seqeralabs/wave/pull/370 first :-)

marcodelapierre:
> LGTM, but is there a way we could avoid hardcoding these in wave, to help keep them in sync with the configs in the stack files?

The effect of these commands in the Dockerfile/Singularityfile is indeed to create the same lines in the referenced Spack YAMLs. 

I get your point on avoiding hard coding. However, I would argue this is quite a good location for them: they are not in the Wave codebase, but instead in a plain text template file. So they are relatively simple to locate, read and edit.  

Note my goal here is not to be necessarily in strict sync with the Spack stacks, as it is not a 100% match anyway (there are differences related to our distinct target audiences, i.e. they also give defaults for MPI packages). It is rather to implement good defaults for ver common base dependencies, to reduce the chances of concretisation/build fails.  See for instance the definitions of `+pic` and `+nls`.  `~llvm` is also a good example: It is a super-lenghty and failure prone build, which (I assume) is not needed for those packages. 

Does these provide good context for you? Happy to take further feedback from you.  
Did you have any specific strategy in mind to avoid hard coding?

bebosudo:
> Did you have any specific strategy in mind to avoid hard coding?

Maybe taking these options as external input? But it's not an urgent fix. Let's deploy these changes first and then we'll edit later if needed.

marcodelapierre:
> Maybe taking these options as external input? But it's not an urgent fix. Let's deploy these changes first and then we'll edit later if needed.

Sounds good Alberto

marcodelapierre:
OK, CI fixed, ready for review @pditommaso

marcodelapierre:
Turning to draft to sort out an issue with the CI commit

marcodelapierre:
@pditommaso  this is now fully ready for review (after CI fix)

src/main/resources/io/seqera/wave/spack/spack-builder-singularityfile.txt
```diff
@@ -32,7 +32,21 @@ EOF
     spack config add concretizer:unify:true
     spack config add concretizer:reuse:true
     spack config add packages:all:target:[{{spack_arch}}]
+    spack -e . config add packages:all:providers:blas:[openblas]
```
pditommaso, line 35: why these uses `spack -e . config add` and the other `spack config add` ?

marcodelapierre: they all need to be with `-e .`, but changes for the other lines are in the sister PR , https://github.com/seqeralabs/wave/pull/368

marcodelapierre: is it better if I merge both PRs into one? this will also make integration testing simpler - test only once

marcodelapierre:
closing , as locally merged into https://github.com/seqeralabs/wave/pull/368

PR #380, opened 2024-01-21
Bump surrealdb/surrealdb:v1.1.0

### LINKED ISSUE #381: Fix support for SurrealDB 1.x
When using the latest stable version of Surreal DB the following test is failing 

https://github.com/seqeralabs/wave/blob/c495d80c1654785627cc5e7337354d8b995a06d2/src/test/groovy/io/seqera/wave/service/persistence/impl/SurrealPersistenceServiceTest.groovy#L289-L324

pditommaso:
This PR bumps surrealdb version v1.1.0


munishchouhan:
surrealdb is throwing an error, when wave tries to insert vulnerabilities
`i.s.w.s.p.i.SurrealPersistenceService - --resv[result:Found 'cve-3' for the id field, but a specific record has been specified, status:ERR, time:21.875µs]`
https://github.com/seqeralabs/wave/blob/c495d80c1654785627cc5e7337354d8b995a06d2/src/main/groovy/io/seqera/wave/service/persistence/impl/SurrealPersistenceService.groovy#L187

src/main/groovy/io/seqera/wave/service/persistence/WaveScanRecord.groovy
```diff
@@ -38,7 +38,7 @@ import io.seqera.wave.util.StringUtils
 @EqualsAndHashCode
 @CompileStatic
 class WaveScanRecord {
-    String id
+    String scanId
```
pditommaso, line 41: What's the rationale of this? I fear this will break all existing records

munishchouhan: now surreal does not allow to have id in record and inside the object too, so we cannot name it id
and I have tested it and everything works fine

munishchouhan: ok correct
the current one in surrealdb will be affected

src/main/groovy/io/seqera/wave/service/persistence/impl/SurrealClient.groovy
```diff
@@ -76,10 +76,10 @@ interface SurrealClient {
     @Put('/key/wave_request/{token}')
     Flux<Map<String, Object>> updateContainerRequestAsync(@Header String authorization, String token, @Body WaveContainerRecord body)
 
-    @Post('/key/wave_scan/{id}')
-    Map<String,Object> insertScanRecord(@Header String authorization, String id, @Body WaveScanRecord body)
+    @Post('/key/wave_scan')
+    Map<String,Object> insertScanRecord(@Header String authorization, @Body WaveScanRecord body)
```
pditommaso, line 80: I'd like to understand why this is not working anymore?

munishchouhan: These API calls, where we are passing id as a separate parameter and id also present inside the object `WaveScanRecord` surrealdb is, throwing error:` result:Found 'cve-3' for the id field, but a specific record has been specified, status:ERR, time:21.875µs`
this is the resposne from surrealdb guys :
Screenshot 2024-01-23 at 19 02 16

https://discord.com/channels/902568124350599239/970338835990974484/1198983764282441809

munishchouhan: I will check here for two scenarios:
1. What id does the assume in the DB when now id is passed in post?
2. What happens when an object with the same id is stored in DB again?

pditommaso:
Along with this, this patch should be not needed anymore. Therefore it could be removed 

https://github.com/seqeralabs/wave/blob/df585bcbb68390f1bf0d1e98deebccf1b92887bf/src/main/groovy/io/seqera/wave/service/persistence/impl/SurrealPersistenceService.groovy#L124-L129

src/main/groovy/io/seqera/wave/service/persistence/impl/SurrealClient.groovy
pditommaso: We may not need to keep compatibility with the old db. See here #387

munishchouhan:
1. What id does the record assume in the DB when no id is passed in post?
Ans:
the is of the record will the tablename:⟨id⟩
and the id will be taken from id inside the object:
Screenshot 2024-01-24 at 16 23 14

munishchouhan:
2. What happens when an object with the same id is stored in DB again?
If there is no change in the object and we do the post again, its works, but if there is a change in the object body then post fails
Screenshot 2024-01-24 at 16 27 36

Incase of change we need to use put:
Screenshot 2024-01-24 at 16 28 05

munishchouhan:
I will add new api calls and also will check if there is a way to do it using sql in a single call

src/main/groovy/io/seqera/wave/service/persistence/impl/SurrealClient.groovy
pditommaso: Look good

pditommaso:
This refactoring looks overkill. The patch of the id should be enough. I've incorporated that change here https://github.com/seqeralabs/wave/pull/388

Issue #387, opened 2024-01-24
Backward compatibility with SurrealDB 1.0-beta8


pditommaso:
Latest versions of SurrealDB introduce some breaking changes with the datastore binary format. 

Unfortunately, the migration process looks to not work properly as well, [see here](https://github.com/surrealdb/surrealdb/issues/2359).

To goal of this issue it allow the use of the latest versions of SurrealDB introduced by https://github.com/seqeralabs/wave/pull/380, without breaking the compatibility with the existing data.

The overall idea consists of using a new DB deployment using the version 1.1.x. that it will be used a primary datastore for any read and write operation, and use the old instance only in read mode, as fallback when a record in not found in the primary DB.

marcodelapierre:
@pditommaso  @munishchouhan  closed by https://github.com/seqeralabs/wave/pull/388  right?

PR #388, opened 2024-01-24
Implement Surreal fallback mechanism to allow migration to new version 

pditommaso:
This PR implements a fallback mechanism to use old version of Surreal in read only mode as described #387


munishchouhan:
CVE can be same for multiple builds, so the post call will fail

pditommaso:
Good point, but the CVE should be the same (ie. immutable) therefore it already exist, it can be assumed that's ok. 

There's a test for this condition 

https://github.com/seqeralabs/wave/blob/ecd4a514ed4dfa0430ca787739aae8e705e85ad0/src/test/groovy/io/seqera/wave/service/persistence/impl/SurrealPersistenceServiceTest.groovy#L303-L308

PR #402, opened 2024-02-14
Added Organisation for seqera platform user

munishchouhan:
This PR will add a new field `String organization` in User

### DIFF
--- a/src/main/groovy/io/seqera/wave/tower/User.groovy
+++ b/src/main/groovy/io/seqera/wave/tower/User.groovy
@@ -29,7 +29,7 @@ import groovy.transform.ToString
  *
  * @author Paolo Di Tommaso <paolo.ditommaso@gmail.com>
  */
-@ToString(includeNames = true, includePackage = false, includes = 'id,userName,email')
+@ToString(includeNames = true, includePackage = false, includes = 'id,userName,email,organization')
 @EqualsAndHashCode
 @CompileStatic
 class User {
@@ -44,4 +44,6 @@ class User {
     @Size(max = 255)
     String email
 
+    @Size(max = 100)
+    String organization
 }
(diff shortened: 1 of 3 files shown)


munishchouhan:
I have tested it and i can see organisation in database now
```
seqera/wave> select user from wave_request where user.id=2 and user.organization!=NONE
[[{ user: { email: 'munish.chouhan@seqera.io', id: 2, organization: 'seqera', userName: 'munish-chouhan' } }, 
{ user: { email: 'munish.chouhan@seqera.io', id: 2, organization: 'seqera', userName: 'munish-chouhan' } }]]
```

src/main/groovy/io/seqera/wave/tower/User.groovy
```diff
@@ -44,4 +44,6 @@ class User {
     @Size(max = 255)
     String email
 
+    @Size(max = 100)
+    String organization
```
pditommaso, line 48: Good try, but this is not the Platform organization but organization name taken from Github when the user sign-in with GH SSO. therefore it's useless. Currently the tower endpoint is the only way to segregate user by org

munishchouhan: this is also the value user set in its Seqera platform profile

pditommaso: Exactly, this is why it's useless

munishchouhan: ok understood

munishchouhan: we already have towerEndpoint in DB
Should I create a metric API for that?

pditommaso: Let put the organization problem on hold for now. If you could start looking the Conda Fatcher PR I've drafted yesterday it could be great

munishchouhan: sure, I will start working on it

pditommaso:
As commented this PR is malformed. The `organization` used not the platform one. Currently is it not possible from the platform API. It should be investigated if it can be inferred via the license manager

PR #404, opened 2024-02-16
Add self-hosted Wave installation guide

jason-seqera:
This PR is `404` ;)

Documentation that describes how to install self-hosted Wave.

Needs more extension instructions on how to configure each of the AWS services, particular IAM roles, but all other services as well.

Going to use `envsubst` for variables.

The formatting of all this is in flight as well, because of the change to `settings.sh` for variable exports.

~The gaps in AWS configuration instructions for services are real, though.~

So this isn't slated to be included with the Wave documentation proper at this time.

### DIFF
--- a/docs/install/index.md
+++ b/docs/install/index.md
@@ -1,0 +1,196 @@
+---
+title: Self-hosted installation
+---
+
+This guide describes how to install the Wave service on your own infrastructure.
+
+## Prerequisites
+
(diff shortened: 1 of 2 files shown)


pditommaso:
Let's keep https://github.com/seqeralabs/wave-k8s-deployment as the k8s install docs for now

PR #415, opened 2024-03-01
Update aws sdk to version v2

munishchouhan:
This PR will update aws SDK to version 2
and also fixed the changes in APIs


src/main/groovy/io/seqera/wave/service/aws/AwsMailProvider.groovy
```diff
@@ -54,15 +59,21 @@ class AwsMailProvider implements MailProvider {
     @Override
     void send(MimeMessage message, Mailer mailer) {
         //get mail client
-        final client = AmazonSimpleEmailServiceClientBuilder
-                .standard()
+        final client = SesClient.builder()
+                .region(Region.of(region))
                 .build()
```
pditommaso, line 64: is it possible to \*not* specify the region?

munishchouhan: This is what I found in docs, if region is not provided then:

    1. Check the 'aws.region' system property for the region.
    2. Check the 'AWS_REGION' environment variable for the region.
    3. Check the {user.home}/.aws/credentials and {user.home}/.aws/config files for the region.
    4. If running in EC2, check the EC2 metadata service for the region.

munishchouhan: we have specified the region in bootstrap.yml, so we can remove it from here
https://github.com/seqeralabs/wave/blob/f8cfdcc17aa2f88cf12fc408f9c3e14c5ed47b78/src/main/resources/bootstrap.yml#L5

munishchouhan: But I think in the case of wave enterprise, we need to provide the option to the user, if the ses is not configured in the same region as their other stuff

pditommaso: It can be assumed that the `AWS_REGION` is going to be available, that's the norm for EKS deployment

munishchouhan: ok sure, I will remove it

build.gradle
```diff
@@ -68,7 +68,7 @@ dependencies {
     implementation("io.micronaut.aws:micronaut-aws-parameter-store")
     implementation "software.amazon.awssdk:ecr"
     implementation "software.amazon.awssdk:ecrpublic"
-    implementation 'com.amazonaws:aws-java-sdk-ses:1.12.471'
+    implementation 'software.amazon.awssdk:ses'
     // include sts to allow the use of service account role - https://stackoverflow.com/a/73306570
     implementation 'com.amazonaws:aws-java-sdk-sts:1.12.471'
```
pditommaso, line 73: also `aws-java-sdk-sts` needs to be updated

munishchouhan: we do have this entry of sts but only for run time
https://github.com/seqeralabs/wave/blob/f17be6b21b674790620202cee0232b92d71ddee6/build.gradle#L79

So I will change this to implementation

pditommaso: Didn't realised about this. In principle `runtimeOnly` should be enough, because it's not used in the implementation

src/test/groovy/io/seqera/wave/controller/ContainerInspectControllerTest.groovy
```diff
@@ -58,18 +58,18 @@ class ContainerInspectControllerTest extends Specification {
         resp.body().container.registry == 'docker.io'
         resp.body().container.imageName == 'library/busybox'
```
pditommaso, line 59: Let's use `library/busybox@sha256:4be429a5fbb2e71ae7958bfa558bc637cf3a61baf40a708cb8fff532b39e52d0` so it will not change overtime

munishchouhan:
@pditommaso  This PR has been tested in Dev, its successful

pditommaso:
Was the email sent tested?

munishchouhan:
> Was the email sent tested?

yes

PR #421, opened 2024-03-12
Refactor build id & status endpoint

pditommaso:
This PR refactor the container build ID to ensure each request has a unique identifier and prevent conflicts for separate requests that may have failed due to a temporary failure. See for example https://github.com/seqeralabs/wave/issues/312.

PR #427, opened 2024-03-21
Registry tests fix

munishchouhan:
Registry controller tests are failing because we are using the latest in tag in those tests, and the latest image has been changed. I have replace latest with sha


pditommaso:
This is already fixed by https://github.com/seqeralabs/wave/pull/421

PR #428, opened 2024-03-21
Fix build restart

munishchouhan:
This PR will add the functionality to restart the build, if it failed last time

### DIFF
--- a/src/main/groovy/io/seqera/wave/service/builder/BuildResult.groovy
+++ b/src/main/groovy/io/seqera/wave/service/builder/BuildResult.groovy
@@ -78,7 +78,7 @@ class BuildResult {
         return "BuildRequest[id=$id; exitStatus=$exitStatus; duration=$duration]"
     }
 
-    static BuildResult completed(String buildId, int exitStatus, String content, Instant startTime, String digest) {
+    static BuildResult completed(String buildId, Integer exitStatus, String content, Instant startTime, String digest) {
         new BuildResult(buildId, exitStatus, content, startTime, Duration.between(startTime, Instant.now()), digest)
     }
 


src/main/groovy/io/seqera/wave/service/builder/ContainerBuildServiceImpl.groovy
```diff
@@ -259,9 +259,14 @@ class ContainerBuildServiceImpl implements ContainerBuildService {
...
         final ret2 = buildStore.getBuild(request.targetImage)
-        if( ret2 ) {
+        if( ret2.getExitStatus() == null  || ret2.getExitStatus() == 0 ) {
             log.info "== Hit build cache for request: $request"
             return new BuildTrack(ret2.id, request.targetImage, true)
+        }else{
+            log.info "== Previous build for this image was failed, building again: $request"
+            buildStore.storeBuild(request.targetImage,ret1)
+            launchAsync(request)
+            return new BuildTrack(ret1.id, request.targetImage, false)
```
pditommaso, line 269: The problem with this approach is that this block is not concurrency safe. This is why above if used `storeIfAbsent` that use an atomic operation to check and update the entry.

pditommaso: I also think is not needed, because when a build fails the entry is stored with a shorter ttl to allow a new retry after a few second. See here 

https://github.com/seqeralabs/wave/blob/d81baa748762a05ceb750bfe5fc2b70be9458688/src/main/groovy/io/seqera/wave/service/builder/ContainerBuildServiceImpl.groovy#L199-L206

munishchouhan: Ok make sense

PR #436, opened 2024-04-01
Fixed RedisTestContainer flaky test

munishchouhan:
This PR will fix the flaky test related to redis container happeing in CI

### DIFF
--- a/src/test/groovy/io/seqera/wave/test/RedisTestContainer.groovy
+++ b/src/test/groovy/io/seqera/wave/test/RedisTestContainer.groovy
@@ -52,6 +52,13 @@ trait RedisTestContainer {
         redisContainer.getMappedPort(6379)
     }
 
+    def setupSpec() {
+        if (!redisContainer.isRunning()) {
+            redisContainer.start()
+            log.debug "Started Redis test container"
+        }
+    }
+
     def cleanupSpec(){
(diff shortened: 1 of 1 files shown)


src/test/groovy/io/seqera/wave/test/RedisTestContainer.groovy
```diff
@@ -52,6 +52,13 @@ trait RedisTestContainer {
         redisContainer.getMappedPort(6379)
     }
 
+    def setupSpec() {
+        if (!redisContainer.isRunning()) {
+            redisContainer.start()
+            log.debug "Started Redis test container"
```
pditommaso, line 58: Wonder if not [all the snippet](https://github.com/seqeralabs/wave/pull/436/files#diff-6bcc5bde499446dc5a43bfc08d5bbd35aeda6055588d2dcb56fb52a417ce3c9bR38-R43) in the static block should be moved here

munishchouhan: We can do that, too, as the point of this change is to make sure that whenever a new test class initiates, it checks whether the Redis is running or not. If it doesn't, then start it. We can go with this; if there are still failures, we can try adding everything in setupspec.

PR #438, opened 2024-04-02
Fix Build pod name

munishchouhan:
This pr will replace _ with - in k8s build pod name to adhere with RFC 1123


pditommaso (changes requested):
Let's add a unit test

pditommaso (approved):

PR #441, opened 2024-04-04
Fix Containerfile encoding

pditommaso:
This PR fixes the Containerfile and Condafile encoding showed in the container view


VERSION
```diff
@@ -1 +1 @@
-1.5.2
+1.6.0-B1
```
```suggestion
1.5.2
```

PR #444, opened 2024-04-08
Fix platform migration

pditommaso:
This PR is to address the issue migrating from tower.nf to cloud.seqera.io


swampie (approved):

VERSION
```diff
@@ -1 +1 @@
-1.6.0-B2
+1.6.0-RC3
```
```suggestion
1.6.0-B2
```

VERSION
```diff
@@ -1 +1 @@
-1.6.0-B2
+1.6.0-B2
```
```suggestion
1.6.0-B2

```

PR #459, opened 2024-04-17
Enable streaming of build logs

### LINKED ISSUE #451: Enable streaming of build logs
The build logs reported in the [build controller](https://github.com/seqeralabs/wave/blob/08f44007f58ae1aba61978554040116313523f4e/src/main/groovy/io/seqera/wave/controller/BuildController.groovy#L63-L72) are only accessible once the logs have been uploaded on the build completion. 

However, it should be possible to fetch the K8s pod output stream to stream the log while the build is running. 

The goal of this issue is to enable this capability

munishchouhan:
This PR will enalbe log streaming to get build log endpoint, while build is in process


src/main/groovy/io/seqera/wave/service/builder/DockerBuildStrategy.groovy
```diff
@@ -113,7 +113,7 @@ class DockerBuildStrategy extends BuildStrategy {
                 ? cmdForKaniko( req.workDir, credsFile, spack, req.platform)
                 : cmdForSingularity( req.workDir, credsFile, spack, req.platform)
 
-        return dockerCmd + launchCmd(req)
+        return dockerCmd + builderName(req.buildId) +launchCmd(req)
```
pditommaso, line 116: Is `builderName` really needed? could not that logic managed by `cmdForKaniko` and `cmdForSingularity` ?

munishchouhan: yes, this is not correct here, this builder name is only need when using docker, I am working on it

munishchouhan:
@pditommaso is this required or should it refresh automatically?

https://github.com/seqeralabs/wave/assets/6575288/8fed1038-a6f7-4ca8-8ad1-ec0bda198900

pditommaso:
If only build box content is refreshed while the status is PENDING, it would be cool

munishchouhan:
> If only build box content is refreshed while the status is PENDING, it would be cool

I will give it a try

munishchouhan:
@pditommaso build logs with refresh now

https://github.com/seqeralabs/wave/assets/6575288/e800b0a6-7092-4707-8ad2-e134ad4231bf

src/main/groovy/io/seqera/wave/service/builder/BuildStrategy.groovy
```diff
@@ -37,6 +37,8 @@ abstract class BuildStrategy {
 
     abstract BuildResult build(BuildRequest req)
 
+    abstract String getLogs(String buildId)
```
pditommaso, line 40: Ideally this should return an `InputStream`

src/main/groovy/io/seqera/wave/service/k8s/K8sServiceImpl.groovy
```diff
@@ -460,6 +462,23 @@ class K8sServiceImpl implements K8sService {
...
+    @Override
+    String getCurrentLogsPod(String name) {
+        try {
+            return k8sClient.coreV1Api().readNamespacedPodLog(name, namespace, null, null, null, null, null, null, null, null, null)
+        } catch (Exception e) {
+            // logging trace here because errors are expected when the pod is not running
+            log.trace "Unable to fetch logs for pod: $name", e
+            return null
+        }
+    }
```
pditommaso, line 480: It should be considered to use a streaming approach, see


https://github.com/kubernetes-client/java/blob/master/examples/examples-release-latest/src/main/java/io/kubernetes/client/examples/LogsExample.java#L36-L49

munishchouhan: I didi tried it, but streamNamespacedPodLog() waits till the pod execution is completed, I will dig more

munishchouhan: I will move the logic to convert string to inputstream to this function so that getLogs return input stream

src/main/groovy/io/seqera/wave/service/builder/BuildStrategy.groovy
munishchouhan: done

munishchouhan:

munishchouhan:
I have moved the docs changes to https://github.com/seqeralabs/wave/pull/464
it will easier to apply once the docs repo is in sync with wave

src/main/groovy/io/seqera/wave/service/k8s/K8sServiceImpl.groovy
pditommaso: I don't know if this is a limitation of java client or a bug, but I'm quite sure it's possible to stream the logs of a running pod. 

I've made a simple test running a pod like this 

```
kubectl run counter --image=busybox --command -- /bin/sh -c 'n=0; while true; do echo $((n++)); sleep 1;
```

Then I've used the [K8s client implemented in Nextflow](https://github.com/nextflow-io/nextflow/blob/1dc4e4e4f043695c413d47fdfda525159b87fd4f/modules/nextflow/src/main/groovy/nextflow/k8s/client/K8sClient.groovy#L571-L582), that's a bare simple http request over URLConnection. 

This snippet just stream the pod logs until the pod completes 

```
    def 'should get pod log' () {
        given:
        def settings = ['context': 'docker-desktop', name: 'default']
        def config = new K8sConfig(settings)
        def client = new K8sClient( config.getClient() )

        when:
        def logs = client.podLog([follow:'true'], 'counter')
        then:
        logs.text == 'hello\n'
    }
```

munishchouhan:
Suggestion from @ewels 
Was thinking about what you were saying about ASCII codes in the logs yesterday 
. I wonder if we can keep them (maybe via a flag in the API?). I would love to be able to render the colours if possible as it makes the logs much easier to read.
The back-end used by Hub is written Python, and my favourite library [Rich](https://github.com/Textualize/rich) is able to convert ascii codes to [html](https://rich.readthedocs.io/en/latest/console.html#exporting). So we might be able to do that on the round trip and preserve colour. On the Wave side we just have to _not_ throw them out. What do you think?

pditommaso:
Please keep colour outside of this PR. We need fist makes the logs stream to work

munishchouhan:
> Please keep colour outside of this PR. We need fist makes the logs stream to work

I will create another issue

munishchouhan:
merged get messed up, moving to draft

Issue #395, opened 2024-02-05
Allow the customisation of container image names


pditommaso:
Wave allow building container images by providing either Dockerfile and Conda packages, and push the resulting image to a repository (not registry) name provided by user e.g. `quay.io/some/repo`. 

When doing so the resulting container images will have the following naming convention when using a dockerfile

```
quay.io/some/repo:<HASH>
```

and  

```
quay.io/some/repo:<CONDA PACKAGE NAME>--<HASH>
```

Where `<HASH>` represents the checksum of the Dockerfile used to build the container. 

The following approach has some drawbacks:

1. when using a dockerfile the resulting image name/tag is still anonymous i.e. there's no way to infer what it does containes 
2. when using more than one Conda package, the resulting tag list all package names becoming difficult to read
3. all contained images are pushed to the same container repository i.e. `quay.io/some/repo` resulting into a collection of heterogeneous packages. 

The goal of this issue is to provide the user the ability to specify the target repository name the should extend the information of the target build repository byt introducing a new `imageName` attribute that would work the following: 

1. `buildRepository: "quay.io/some/repo"` and `imageName: <empty>` uses the current behavior 
2. `buildRepository: "quay.io/some/repo"` and `imageName: "foo"` results in the container name `quay.io/some/repo/foo:<HASH>`
3. `buildRepository: "quay.io/some/repo"` and `imageName: "foo/bar"` results in the container name `quay.io/some/repo/foo/bar:<HASH>`
4.  `buildRepository: "quay.io"` and `imageName: "foo/bar"` results in the container name `quay.io/foo/bar:<HASH>`

Possible drawbacks: 

* A malicious user could use this approach, especially 4., to override container images in arbitrary  repository

munishchouhan:
@pditommaso should I start working on it?

ewels:
Great! I think that this will be really helpful, I much prefer having separate docker repos for different software tools rather than one for everything..

> A malicious user could use this approach, especially 4., to override container images in arbitrary repository

I don't entirely follow this. For Wave to push an image to eg. quay.io the use has to have access credentials. If they have credentials then they can override container images already, no? Or do you mean that this could lead to people doing this by accident?

ewels:
The public registry is a special case, so we may need some special behaviour here to control scoping above the root level.

munishchouhan:
Three PRs have been created to achieve this in Wave.

1. https://github.com/seqeralabs/libseqera/pull/13
[...]

PR #460, opened 2024-04-18
Improve build container naming strategy

pditommaso:
Implements container naming strategy as described [here](https://github.com/seqeralabs/wave/issues/395#issuecomment-2059492650)


src/test/groovy/io/seqera/wave/core/RoutePathTest.groovy
```diff
@@ -136,4 +136,6 @@ class RoutePathTest extends Specification {
         route2.request.containerImage == 'ubuntu:latest'
         route2.identity == new PlatformId(new User(id: 100))
     }
+
+
```
pditommaso, line 140:
```suggestion
```

PR #465, opened 2024-04-22
added null check for container image name

munishchouhan:
This PR will fix the NPE thrown by parse method in ContainerCoordinates when path is not specified


pditommaso (changes requested):
Worth adding also a BadRequestException in the controller entrypoint

munishchouhan:
tested in local:
Screenshot 2024-04-22 at 15 24 51

pditommaso (approved):

Issue #469, opened 2024-04-23
Aggregate metrics


pditommaso:
The goal of this issue is to extend the v1alpha2 metrics endpoint so that it returns aggregate metrics when no `date` or `org` parameter is specified. 

In this case, it should return the value of all `<metrics>/o/<organization>` keys and the total for all for them for example having 

```
/pulls/o/foo: 1
/pulls/o/bar: 5
```

it should return 

```
{
  "metrics": "pulls",
  "count": 6,
  "orgs": {
    "foo": 1,
    "bar": 5
  }
}

PR #472, opened 2024-04-24
Update metric service to get aggregated metrics 

munishchouhan:
closes https://github.com/seqeralabs/wave/issues/469
1. This PR will add a function in counterProvider to get all matching data data using hscan
2. This will add a method in metric service to get the count per org and total count
3. Add new model for the response

When no date or ord is provided then metrics endpoint will return the above

### DIFF
--- a/src/main/groovy/io/seqera/wave/controller/MetricsController.groovy
+++ b/src/main/groovy/io/seqera/wave/controller/MetricsController.groovy
@@ -18,6 +18,7 @@
 
 package io.seqera.wave.controller
 
+import java.util.regex.Pattern
 import javax.annotation.Nullable
 
 import groovy.transform.CompileStatic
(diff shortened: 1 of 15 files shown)


munishchouhan:
Tested successfully in local
Screenshot 2024-04-24 at 16 34 38
Screenshot 2024-04-24 at 16 44 27

Screenshot 2024-04-24 at 16 42 28

munishchouhan:
Working on only getting the /metrics/o records from redis

munishchouhan:
> Working on only getting the /metrics/o records from redis

Hscan uses glob pattern, so we need to get all the records in metrics/o/ and then filter out the relevant ones

munishchouhan:

pditommaso:
Got an error trying this 

```
» curl -u foo:bar http://localhost:9090/v1alpha2/metrics/fusion/pulls
{"message":"Cannot invoke \"java.util.Map.entrySet()\" because \"self\" is null - Error ID: 375c044c6d5f"}
```

munishchouhan:
> Got an error trying this
> 
> ```
> » curl -u foo:bar http://localhost:9090/v1alpha2/metrics/fusion/pulls
> {"message":"Cannot invoke \"java.util.Map.entrySet()\" because \"self\" is null - Error ID: 375c044c6d5f"}
> ```

fixed

pditommaso:
Ok, nice

PR #474, opened 2024-04-26
Ignore order of packages and channels in build requests

vladsavelyev:
Fix https://github.com/seqeralabs/wave/issues/473


pditommaso:
Thanks Vlad, but don't think Wave should do this, otherwise should also re-order packages in a conda yaml file.

vladsavelyev:
No problem - hub can do that instead 👍

PR #482, opened 2024-04-30
Fix ci errors

munishchouhan:
This PR will make deletion of workspace at test failure in CI

verified
Screenshot 2024-04-30 at 17 37 01


munishchouhan:
@pditommaso this is a very small change, please approve it
it will help to get the reports when test fails

pditommaso:
It will fail .. 😆

munishchouhan:
> It will fail .. 😆

if it fails, we will know if it works or not 😆

PR #487, opened 2024-05-03
Fix aggregate metrics issue 

munishchouhan:
This PR will add count to force redis to scan certain number of records

### DIFF
--- a/src/main/groovy/io/seqera/wave/service/counter/impl/RedisCounterProvider.groovy
+++ b/src/main/groovy/io/seqera/wave/service/counter/impl/RedisCounterProvider.groovy
@@ -55,7 +55,7 @@ class RedisCounterProvider implements CounterProvider {
     @Override
     Map<String, Long> getAllMatchingEntries(String key, String pattern) {
         try(Jedis conn=pool.getResource() ) {
-            final scanResult = conn.hscan(key, "0", new ScanParams().match(pattern))
+            final scanResult = conn.hscan(key, "0", new ScanParams().match(pattern).count(10000))
             if( !scanResult )
                 return Map.<String, Long>of()
             final result = new HashMap<String, Long>()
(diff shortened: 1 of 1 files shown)


src/main/groovy/io/seqera/wave/service/counter/impl/RedisCounterProvider.groovy
```diff
@@ -55,7 +59,7 @@ class RedisCounterProvider implements CounterProvider {
     @Override
     Map<String, Long> getAllMatchingEntries(String key, String pattern) {
         try(Jedis conn=pool.getResource() ) {
-            final scanResult = conn.hscan(key, "0", new ScanParams().match(pattern).count(10000))
+            final scanResult = conn.hscan(key, "0", new ScanParams().match(pattern).count(hscanCount))
```
pditommaso, line 62: what's the rationale?

munishchouhan: so that it can be changed, if we want to scan more records than the default amount

munishchouhan: official docs of count
https://redis.io/docs/latest/commands/scan/
```
The COUNT option

While SCAN does not provide guarantees about the number of elements returned at every iteration, it is possible to empirically adjust the behavior of SCAN using the COUNT option. Basically with COUNT the user specifies the amount of work that should be done at every call in order to retrieve elements from the collection. This is just a hint for the implementation, however generally speaking this is what you could expect most of the times from the implementation.

    The default COUNT value is 10.
    When iterating the key space, or a Set, Hash or Sorted Set that is big enough to be represented by a hash table, assuming no MATCH option is used, the server will usually return count or a few more than count elements per call. Please check the why SCAN may return all the elements at once section later in this document.
    When iterating Sets encoded as intsets (small sets composed of just integers), or Hashes and Sorted Sets encoded as ziplists (small hashes and sets composed of small individual values), usually all the elements are returned in the first SCAN call regardless of the COUNT value.
```

pditommaso (approved):

PR #494, opened 2024-05-10
Improve thread pool handling

pditommaso:
This PR improves the handling of 'future-store-executor' thread pool, offloading the execution to the 'io-executor' thread pool


VERSION
```diff
@@ -1 +1 @@
-1.7.5
+1.7.6-B1
```
```suggestion
1.7.5
```

Issue #498, opened 2024-05-13
Prevent the use of Community registry with containerImage attribute


pditommaso:
The use of Community registry and freeze option should only be allow when specifying the `packages` attribute. 

The following request should not be allowed 

```
{
  "containerImage": "alpine",
  "dryRun": false,
  "freeze": true,
  "timestamp": "2024-05-13T13:10:55.291206+02:00",
  "towerAccessToken": "<omitted>",
  "towerEndpoint": "<omitted>"
}
```

pditommaso:
Solved by #499

PR #499, opened 2024-05-13
Prevent container freeze without custom repo

pditommaso:
This PR prevents the request for freeze container without a custom build repository. Solves  #498


VERSION
```diff
@@ -1 +1 @@
-1.7.7
+1.7.8-B2
```
```suggestion
1.7.7
```

PR #503, opened 2024-05-16
Replace kaniko with buildkit

munishchouhan:
This PR will replace the builder image with Buildkit. 

Solves #478 and #287.

PR #510, opened 2024-05-23
removed unused imports

munishchouhan:
This PR will do the following:
1. Remove unused imports in build request class
2. ~~remove .trivy-cache in build ci~~

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,9 +72,9 @@ jobs:
           AZURECR_PAT: ${{ secrets.AZURECR_PAT }}
           GOOGLECR_KEYS: ${{ secrets.GOOGLECR_KEYS }}
 
-      - name: Cleanup build workspace
+      - name: Cleanup trivy cache
         run: |
-          sudo rm -rf /home/runner/work/wave/wave/build-workspace
+          sudo rm -rf /home/runner/work/wave/wave/build-workspace/.trivy-cache
 
       - name: Release
         if: "contains(github.event.head_commit.message, '[release]')"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -75,7 +75,7 @@ jobs:
       - name: Cleanup build workspace
         if: always()
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -72,9 +72,9 @@ jobs:
           AZURECR_PAT: ${{ secrets.AZURECR_PAT }}
           GOOGLECR_KEYS: ${{ secrets.GOOGLECR_KEYS }}
 
-      - name: Cleanup build workspace
+      - name: Cleanup trivy cache
         run: |
-          sudo rm -rf /home/runner/work/wave/wave/build-workspace
+          sudo rm -rf /home/runner/work/wave/wave/build-workspace/.trivy-cache
```
pditommaso, line 77: Is this problem still around? why only .trivy-cache ?

munishchouhan: yes, the problem is still around, I think in the previous command its not delete the hidden folders, and if i use .*, it does not execute, the problem is related to the access of `.trivy-cache`. So i tried it in this PR https://github.com/seqeralabs/wave/pull/503 and ci start working 
Screenshot 2024-05-23 at 15 54 11

munishchouhan: previous tries

1. with `sudo rm -rf /home/runner/work/wave/wave/build-workspace/.*`
Screenshot 2024-05-23 at 15 56 57

3. with `sudo rm -rf /home/runner/work/wave/wave`
Screenshot 2024-05-23 at 15 57 19

pditommaso: I'm quite sure `rm -rf /foo/bar` remove ALL bar content and the directory itself. therefore it would solve

munishchouhan: you are right, the problem is github ci skip this delete step in some failures, i am not sure why 
as happened in this run
https://github.com/seqeralabs/wave/actions/runs/9206439473/job/25324299024

pditommaso: Likely [this](https://github.com/seqeralabs/wave/commit/8adb52ab355dd72ff3d4d19a51d42a63fdc2e546) should solve

PR #544, opened 2024-06-25
API to search build details

### LINKED ISSUE #541: Endpoint to look up build id by container name
Create API endpoint to get build details against the provided container name

munishchouhan:
This PR will add an API to get builds for a specific image or user

API: `/v1alpha1/builds`
query parameters: `imageName` and `user`


munishchouhan:
@ewels In this PR I am fetching records only from database.
Now, when a build starts, Wave creates a record in the cache first, and when the build completes, it saves the record in the database and also updates it in the cache. So, do you need records from the cache, too?

pditommaso:
Closing as the view endpoint now allows retrieving the build info with a partial buildId (ie. using the containerId in the image tag)

PR #558, opened 2024-07-08
Update buildlogs aws config to not initialize AwsS3Operations

munishchouhan:
This PR will update buildlogs aws config  to not use micronaut default config

### DIFF
--- a/src/main/resources/application-buildlogs-aws.yml
+++ b/src/main/resources/application-buildlogs-aws.yml
@@ -1,7 +1,7 @@
 ---
-micronaut:
-  object-storage:
-    aws:
-      build-logs:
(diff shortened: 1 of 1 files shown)


pditommaso:
Think it should removed also this section 

https://github.com/seqeralabs/wave/blob/f27ff52766aa82acb715cc6b0f7946819b9bcd40/src/main/resources/application-buildlogs-aws-test.yml#L2-L7

and this 

https://github.com/seqeralabs/wave/blob/f27ff52766aa82acb715cc6b0f7946819b9bcd40/src/main/resources/application-buildlogs-local.yml#L8-L15

pditommaso:
Ok, i've removed them

PR #574, opened 2024-07-24
Extend build timeout

pditommaso:
This PR mitigates the issue with build timeout when the build cluster in congestioned. 

The overall idea is to double the overall build await the time, so the additional waiting scheduling time is taken into account before erroring the job build. 

This may be a stopgap solution. Another approach to be investigated is to use to extend the `BuildStrategy` to a method to check the current state of the build pod/container.


munishchouhan:
I found this about `activeDeadlineSeconds` in https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#detailed-behavior

`Use activeDeadlineSeconds on the Pod to prevent init containers from failing forever. The active deadline includes init containers. However it is recommended to use activeDeadlineSeconds only if teams deploy their application as a Job, because activeDeadlineSeconds has an effect even after initContainer finished. The Pod which is already running correctly would be killed by activeDeadlineSeconds if you set.`

so should we increase timeout for pod too?

pditommaso:
Since there's no use of "init container" it looks not relevant to me

PR #576, opened 2024-07-24
Add delete after finish to blob cache jobs

pditommaso:
This PR adds the ability to delete blob cache jobs automatically by using the `ttlSecondsAfterFinished` job spec. 

By default it used 10 days.


pditommaso:
@munishchouhan can you please add a test?

munishchouhan:
> @munishchouhan can you please add a test?

done

PR #581, opened 2024-07-25
Improve credentials error message consider docker.io as default repo

munishchouhan:
This PR will add a condition in creds check for docker.io


pditommaso:
Close in favour of #665 that uses common parsing logic

PR #596, opened 2024-08-08
Improve blob cache reliability

pditommaso:
This PR improves the Blob caching mechanism reliability in two ways: 

1. checking directly the existence of a cached blob in the target, because the local status can be stale 
2. fetching the status of the transfer job from the currently underlying platform to prevent a timeout event when the status in the redis cache is staled


munishchouhan:
status is specific fo kubestrategy
so can i make status() as Default method and override only in KubeTransferStrategy?

munishchouhan:
I have tested it locally using k8s, where i retarded the application in the mid of blob upload, after restart the pull resumes successfully

munishchouhan:
I will make some more tests

src/main/groovy/io/seqera/wave/service/blob/impl/SimpleTransferStrategy.groovy
```diff
@@ -55,4 +55,11 @@ class SimpleTransferStrategy implements TransferStrategy {
         builder.redirectErrorStream(true)
         return builder
     }
+
+    @Override
+    Status status(BlobCacheInfo info) {
+        // TODO
+        return null
```
pditommaso, line 62: not easy this one 😓

src/main/groovy/io/seqera/wave/service/k8s/K8sService.groovy
```diff
@@ -33,7 +33,7 @@ import io.seqera.wave.configuration.SpackConfig
  */
 interface K8sService {
 
-    enum JobStatus { Pending, Running, Succeeded, Failed }
+    enum JobStatus { PENDING, RUNNING, SUCCEED, FAILED, UNKNOWN }
```
pditommaso, line 36: Can the `JobStatus` remain as it was before?

src/main/groovy/io/seqera/wave/service/blob/impl/SimpleTransferStrategy.groovy
pditommaso: I've found a nice trick with chatgpt 

```
# Set a unique process name using prctl
exec -a "my_unique_process_name" bash -c 'while true; do sleep 60; done'
```

pditommaso: this can be used to change the process name, then it's possible to find the process pid with 

```
ps -eo pid,comm | grep 'my_unique_process_name' | grep -v grep | awk '{print $1}'
```

pditommaso: tho this will tell only if it's running or not running

src/main/groovy/io/seqera/wave/service/k8s/K8sService.groovy
munishchouhan: currently job is only used by transfer, so here we have opportunity to standardize it and make it same as you did for transfer. But If you want to keep the old one i can add mapping between job and transfer status

pditommaso: > But If you want to keep the old one i can add mapping between job and transfer status

Yes, let's not mix different things

src/main/groovy/io/seqera/wave/service/blob/impl/SimpleTransferStrategy.groovy
munishchouhan: go this error 
```
java.io.IOException: Cannot run program "exec": error=2, No such file or directory
	at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1143)
	at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1073)
	at io.seqera.wave.service.blob.impl.SimpleTransferStrategy.transfer(SimpleTransferStrategy.groovy:44)
```

munishchouhan: not getting error anymore, for this, i wasn't adding sh -c in front

munishchouhan: if i do like this
`sh -c exec -a  <jobname> curl -s -X GET ......<Rest>`
it immediately completes without uploading anything

munishchouhan: found the problem, exec works with bash, changes to `bash -c exec -a  <jobname> curl -s -X GET ......<Rest>`

pditommaso:
A couple of things to finalise this PR 

1. let's add the `transfer-` prefix to the `BlobCacheInfo.id` so the `jobName` is not needed anymore
2. let's add unit tests for new methods

munishchouhan:
I found one issue, when application restarts, it creates new jobs again for the previous transfers

pditommaso:
What do you mean? a k8s job?

munishchouhan:
ok I tink the issue is here, previously we were using targetpath in cache, which remains same, but id will change with every request even if the transfer is same
https://github.com/seqeralabs/wave/blob/68b28b674c0d0b7e44ef77e50f74172bf9fe294f/src/main/groovy/io/seqera/wave/service/blob/impl/BlobCacheServiceImpl.groovy#L105

munishchouhan:
> What do you mean? a k8s job?

I mean new k8s jobs getting created, if there is ongoing blob upload and app restarts

munishchouhan:
if we chnage back to targetPath, we will get this error:
https://github.com/seqeralabs/wave/blob/68b28b674c0d0b7e44ef77e50f74172bf9fe294f/src/main/groovy/io/seqera/wave/service/blob/transfer/TransferManager.groovy#L99

munishchouhan:
either we need to make id same of every same blob request
or we can use targetpath in tranferqueue also

pditommaso:
> but id will change with every request even if the transfer is same

That's because the `id` is check summed with current instant. That was done to have a unique job id.
 
https://github.com/seqeralabs/wave/blob/3da7961f79089d10a64accc8c4475f6e5e3b246e/src/main/groovy/io/seqera/wave/service/blob/BlobCacheInfo.groovy#L207-L214

munishchouhan:
I have tested it with 10 parallel pulls and started the application multiple times 
all images downloaded successfully

munishchouhan:
I will test it in dev now

munishchouhan:
Tested in dev successfully

munishchouhan:

src/main/groovy/io/seqera/wave/service/blob/BlobCacheInfo.groovy
```diff
@@ -98,14 +109,16 @@ class BlobCacheInfo {
...
+    static BlobCacheInfo create(String locationUri, String objectUri, Map<String,List<String>> request, Map<String,List<String>> response) {
         final headers0 = new LinkedHashMap<String,String>()
         for( Map.Entry<String,List<String>> it : request )
             headers0.put( it.key, it.value.join(',') )
         final length = headerLong0(response, 'Content-Length')
         final type = headerString0(response, 'Content-Type')
         final cache = headerString0(response, 'Cache-Control')
-        new BlobCacheInfo(locationUrl, headers0, length, type, cache, Instant.now(), null, null, null)
+        final creationTime = Instant.now()
+        new BlobCacheInfo(generateId(objectUri), locationUri, objectUri, headers0, length, type, cache, creationTime,
```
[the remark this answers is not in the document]
munishchouhan: done

src/main/groovy/io/seqera/wave/service/blob/BlobCacheInfo.groovy
```diff
@@ -84,6 +90,11 @@ class BlobCacheInfo {
      */
     final String logs
 
+    /**
+     * it is the name of k8s job or docker container depends on the transfer strategy
+     */
+    final String jobName
```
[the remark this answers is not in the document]
munishchouhan: done

src/main/groovy/io/seqera/wave/service/blob/impl/DockerTransferStrategy.groovy
```diff
@@ -36,28 +39,29 @@ import jakarta.inject.Inject
...
-    BlobCacheInfo transfer(BlobCacheInfo info, List<String> command) {
-        final proc = createProcess(command).start()
-        // wait for the completion and save the result
-        final completed = proc.waitFor(blobConfig.transferTimeout.toSeconds(), TimeUnit.SECONDS)
-        final int status = completed ? proc.exitValue() : -1
-        final logs = proc.inputStream.text
-        return info.completed(status, logs)
+    void transfer(BlobCacheInfo info, List<String> command) {
+        // create a unique name for the container
+        createProcess(command, info.id, blobConfig.transferTimeout.toSeconds()).start()
```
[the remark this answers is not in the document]
munishchouhan: done

src/main/groovy/io/seqera/wave/service/blob/impl/DockerTransferStrategy.groovy
```diff
@@ -73,4 +77,81 @@ class DockerTransferStrategy implements TransferStrategy {
         builder.redirectErrorStream(true)
         return builder
     }
+
+    @Override
+    Transfer status(BlobCacheInfo blob) {
+        final state = getDockerContainerState(blob.id)
```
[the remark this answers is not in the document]
munishchouhan: done

src/main/groovy/io/seqera/wave/service/blob/impl/KubeTransferStrategy.groovy
```diff
@@ -61,63 +61,59 @@ class KubeTransferStrategy implements TransferStrategy {
...
+        // determine exit code and logs
+        final exitCode = pod
+                .status
+                ?.containerStatuses
+                ?.first()
+                ?.state
+                ?.terminated
+                ?.exitCode
+        final stdout = k8sService.logsPod(pod)
+        return new Transfer(mapToStatus(status), exitCode, stdout)
```
pditommaso, line 97: Considering the pod run by the job is expected to be terminated, i've tried to simplify this part. @munishchouhan can you please review and test it?

munishchouhan: ok sure

munishchouhan:
Tested successfully in local with k8s and dev blob cache

munishchouhan (approved):

munishchouhan:
Tested successfully with docker too

VERSION
```diff
@@ -1 +1 @@
-1.10.4
+1.11.0-B0
```
```suggestion
1.10.4
```

PR #598, opened 2024-08-12
Blob cache improve v2

pditommaso:
This PR reorganise the handling of blob cache upload to make it reliable across restarted. 

Main changes:
1. do not block waiting for the job completion when a transfer is launched 
2. implement the (K8s/docker) job monitoring async in a separate service named [TransferManager](https://github.com/seqeralabs/wave/blob/fb3309bfbab0d0b6a0e2b540bb14a1270a950ba9/src/main/groovy/io/seqera/wave/service/blob/transfer/TransferManager.groovy#L45-L45)
3. Each replica runs its own transfer manager instance 
4. The transfer manager pick the id/name of the job to check from the [transfer queue](https://github.com/seqeralabs/wave/blob/fb3309bfbab0d0b6a0e2b540bb14a1270a950ba9/src/main/groovy/io/seqera/wave/service/blob/transfer/TransferManager.groovy#L57-L57) persisted by a Redis list 
5. The blob status should be updated/written by the Transfer manager
6. Removed the S3 file size check, because very likely is useless 
7. Removed `SimpleTransferStrategy` to simplify the implementation


src/main/groovy/io/seqera/wave/service/blob/impl/BlobCacheServiceImpl.groovy
```diff
@@ -193,55 +178,20 @@ class BlobCacheServiceImpl implements BlobCacheService {
...
-            return result
-        }
-    }
-
-    /**
-     * Check the size of the blob stored in the cache
-     *
-     * @return {@link BlobCacheInfo} the blob cache info
-     */
-    protected BlobCacheInfo checkUploadedBlobSize(BlobCacheInfo info, RoutePath route) {
```
munishchouhan, line 221: no blob size checking anymore?

pditommaso: This is useless. Both S3 and R2 implements strong consistency (upload cannot be partial)

https://developers.cloudflare.com/r2/reference/consistency/#operations-and-consistency

munishchouhan: ok understood

munishchouhan:
Tested in Local using k8s and dev blob cache
I restarted the application in between the upload and docker pull resumed successfully after that

src/main/groovy/io/seqera/wave/service/blob/transfer/TransferQueue.groovy
```diff
@@ -45,4 +45,8 @@ class TransferQueue {
         transferQueue.poll(QUEUE_NAME, timeout)
     }
 
+    String pollAndGetValue(Duration timeout){
+        def transferEntry = poll(timeout)
+        return transferEntry ? transferEntry.substring(transferEntry.indexOf(QUEUE_NAME) + QUEUE_NAME.length() + 1) : null
```
pditommaso, line 50: This is a bit messy. Think we need to unify `BlobCacheInfo.id` and `BlobCacheInfo.jobName`

munishchouhan: sorry, I don't understand
the poll is returning transfer-queue/v1=<id>
and we only need id here
https://github.com/seqeralabs/wave/blob/401638800bf968737968473d1298234a0a942593/src/main/groovy/io/seqera/wave/service/blob/transfer/TransferManager.groovy#L79

pditommaso: Ohh, I was misunderstanding it. I didn't realised it because i've tested with the "local" queue. Then it should handle both cases and please add a unit test

pditommaso: But if so, why this is OK ? 

https://github.com/seqeralabs/wave/blob/37174aeab5557686141fe9be8095d16be31fa49b/src/test/groovy/io/seqera/wave/service/data/queue/RedisMessageQueueTest.groovy#L66-L77

munishchouhan: Yes, I have noticed that. I am debugging this right now
Screenshot 2024-08-12 at 14 52 01

munishchouhan: I have updated the poll method to return just the value

src/main/groovy/io/seqera/wave/service/data/queue/impl/RedisMessageQueue.groovy
```diff
@@ -61,7 +61,8 @@ class RedisMessageQueue implements MessageQueue<String>  {
     String poll(String target, Duration duration) {
         try (Jedis conn = pool.getResource()) {
             double d = duration.toMillis() / 1000.0
-            return conn.brpop(d, target)
+            def keyValue =  conn.brpop(d, target)
+            return keyValue != null ? keyValue.getValue() : null
```
pditommaso: Interesting, let's add a test to RedisMessageQueue then

Issue #602, opened 2024-08-18
Refactor `TransferManager` into `JobManager`


pditommaso:
The [TransferManager](https://github.com/seqeralabs/wave/blob/a09980cf6c1bee3fc192af541562ea70ec82e0b3/src/main/groovy/io/seqera/wave/service/blob/transfer/TransferManager.groovy#L47-L47) and [TransferQueue](https://github.com/seqeralabs/wave/blob/a09980cf6c1bee3fc192af541562ea70ec82e0b3/src/main/groovy/io/seqera/wave/service/blob/transfer/TransferQueue.groovy#L33-L33) classes are used to handle the async execution of transfer jobs (ie. upload) for caching blob layer. 

However there are multiple use cases in which this pattern can be used, e.g. [build and scan containers](https://github.com/seqeralabs/wave/issues/575), [SBOM generation](https://github.com/seqeralabs/wave/issues/489), etc.  

The goal of this issues is to refactor `TransferManager` in a `JobManager` for the handling of generic jobs. In a similar vein `TransferQueue` should be refactor in `JobQueue` to handle generic job IDs. 

The job queue should should hold entries defined as 

```
class JobRequest {
  String id
  Type type

  enum Type { Transfer, Build, Scan } 
}
```

pditommaso:
Solved by #575, #605, #623 and #624

PR #605, opened 2024-08-18
Add Job manager 

pditommaso:
This PR adds a job manager to handle the async execution of generic long running jobs. 

### Context 

It needs the ability to submit the execution of jobs via queue shared across multiple replicas. Each replica can pick a job and offload the execution by using a Docker container or Kubernetes job. 

### Requirement 

* The queue needs to be persisted so that the jobs are not impacted by backend restarted 
* It must be able to support multiple running backend instances 
* The same job must be executed one and only one time. 

### Solution 

Use Redis Stream to guarantee persistence, uniqueness and consistency across restarts and scaling events. Key points:
1. When a job needs to be executed is added to a Redis stream using [XADD](https://github.com/seqeralabs/wave/blob/73abea0c553c6ba27383007aff909b49021c3162/src/main/groovy/io/seqera/wave/service/data/stream/impl/RedisMessageStream.groovy#L99-L99) operation 
2. Multiple consumers read the stream by using [XREADGROUP](https://github.com/seqeralabs/wave/blob/73abea0c553c6ba27383007aff909b49021c3162/src/main/groovy/io/seqera/wave/service/data/stream/impl/RedisMessageStream.groovy#L125-L125)
3. The semantic of `XREADGROUP` guarantees that only one consumer can read the same job from the stream 
4. Once the job execution is complete the operation is confirmed by `XACK`, removing permanently the job form the Redis stream. 
5. Both in the case the job is still running or of a sudden system crash, the `XACK` is not carried out. This causes (another) consumer to be able to read again the same job from the stream by using the [XAUTOCLAIM](https://github.com/seqeralabs/wave/blob/73abea0c553c6ba27383007aff909b49021c3162/src/main/groovy/io/seqera/wave/service/data/stream/impl/RedisMessageStream.groovy#L143-L143) command the claim interval as expired  

Solves #602


VERSION
```diff
@@ -1 +1 @@
-1.10.4
+1.11.0-B4
```
```suggestion
1.10.4
```

pditommaso:
This has been rewritten from scratch using Redis streams. The use of [xclaim](https://github.com/seqeralabs/wave/blob/73abea0c553c6ba27383007aff909b49021c3162/src/main/groovy/io/seqera/wave/service/data/stream/impl/RedisMessageStream.groovy#L138-L156) should protected against unexpected interruptions. 

@munishchouhan give it a try at your convenience. 

Could it be useful also a review by @t0randr of the [RedisMessageStream](https://github.com/seqeralabs/wave/blob/73abea0c553c6ba27383007aff909b49021c3162/src/main/groovy/io/seqera/wave/service/data/stream/impl/RedisMessageStream.groovy#L52-L52) implementation

t0randr:
> This has been rewritten from scratch using Redis streams. The use of [xclaim](https://github.com/seqeralabs/wave/blob/73abea0c553c6ba27383007aff909b49021c3162/src/main/groovy/io/seqera/wave/service/data/stream/impl/RedisMessageStream.groovy#L138-L156) should protected against unexpected interruptions.
> 
> @munishchouhan give it a try at your convenience.
> 
> Could it be useful also a review by @t0randr of the [RedisMessageStream](https://github.com/seqeralabs/wave/blob/73abea0c553c6ba27383007aff909b49021c3162/src/main/groovy/io/seqera/wave/service/data/stream/impl/RedisMessageStream.groovy#L52-L52) implementation

The redis implementation seems ok to me.

On a minor note I personally find a bit confusing that we use a `Predicate` as the handler for `consume`.

munishchouhan:
tested successfully locally with 10 parallel image pulls with and without redis
Will test in dev

munishchouhan:
Tested successfully in dev with 10 parallel pulls

src/main/groovy/io/seqera/wave/service/k8s/K8sServiceImpl.groovy
```diff
@@ -585,7 +585,7 @@ class K8sServiceImpl implements K8sService {
      *      The {@link V1Job} description the submitted job
      */
     @Override
-    V1Job transferJob(String name, String containerImage, List<String> args, BlobCacheConfig blobConfig) {
+    V1Job launchJob(String name, String containerImage, List<String> args, BlobCacheConfig blobConfig) {
```
munishchouhan, line 588: maybe in the separate PR, but this method needs to be refactored to be generic for all the jobs, once scan and build are implemented, we can have separate  spec methods and can be invoked by check the type, but this method needs to be generic

pditommaso: Agree. Let's keep into next PR

pditommaso:
> n a minor note I personally find a bit confusing that we use a Predicate as the handler for consume

True, that's because initially it was a `Consumer` instead of a `Predicate`, but at the same time `consume` expressed better that consumes the content of the stream. Predicate is a bit meaningless in this context.

Issue #603, opened 2024-08-18
Extend BuildStatus with RUNNING state 


pditommaso:
The [BuildStatus](https://github.com/seqeralabs/wave/issues/602) information should be extended to report the transition from pending to running state. 

Requires https://github.com/seqeralabs/wave/issues/602

PR #607, opened 2024-08-20
add build in progress status in build page

munishchouhan:
This pr will add a binding in build view to show in progress status
Screenshot 2024-08-26 at 12 06 31


pditommaso:
Nice. I think we used grey in other pages? may the scan view? can you please align the style to it?

pditommaso (changes requested):
Nice. I think we used grey in other pages? may the scan view? can you please align the style to it?

munishchouhan:
> Nice. I think we used grey in other pages? may the scan view? can you please align the style to it?

@pditommaso are you referring to the background color?

munishchouhan:
style has been synced with scan
Screenshot 2024-08-23 at 13 18 11

pditommaso:
great!

ewels:
Can we align with Platform? That uses a blue for in-progress. Really I think that the design team should be involved here..

pditommaso:
Desirable bot not a goal of this PR

src/main/groovy/io/seqera/wave/controller/ViewController.groovy
```diff
@@ -75,6 +75,8 @@ class ViewController {
         final binding = new HashMap(20)
         binding.build_id = result.buildId
         binding.build_success = result.succeeded()
+        binding.build_failed = result.exitStatus  && result.exitStatus != 0
+        binding.build_in_progress = result.exitStatus == null
         binding.build_exit_status = result.exitStatus
```
pditommaso, line 80: Considering we are going to add also a [running](https://github.com/seqeralabs/wave/issues/603) status, it could make sense to encapsulate this logic into `WaveBuildRecord` adding a `Status getStatus()` method

munishchouhan: done

src/main/groovy/io/seqera/wave/controller/ViewController.groovy
```diff
@@ -74,8 +75,17 @@ class ViewController {
...
+        switch (result.status) {
+            case BuildStatus.SUCCESS:
+                binding.build_success = true
+                break
+            case BuildStatus.FAILURE:
+                binding.build_failed = true
+                break
+            case BuildStatus.RUNNING:
+                binding.build_in_progress = true
+        }
```
pditommaso, line 88: This likely can be simplified using `build_state` instead of multiple flags

pditommaso (changes requested):

src/main/groovy/io/seqera/wave/controller/ViewController.groovy
munishchouhan: HBS does not support compare between values, that's why i have used it 
https://handlebarsjs.com/guide/builtin-helpers.html#if

pditommaso: ohh, didn't realised that. Than does not make much sense.

pditommaso:
How hard would be to make the page refresh while the status is "in progress" ? think it has been done somewhere else

pditommaso:
Likely this https://github.com/seqeralabs/wave/blob/389966f271c7198567cb85e5b912073269887281/src/main/resources/io/seqera/wave/scan-view.hbs#L4-L6

munishchouhan:
> How hard would be to make the page refresh while the status is "in progress" ? think it has been done somewhere else

I will add it

munishchouhan:
refresh has been added and tested successfully

PR #612, opened 2024-08-21
Fix outdated Platform docs URLs on nextflow.mdx

llewellyn-sl:


jason-seqera:
@llewellyn-sl, as it turns out it isn't possible to cross-link this way. This needs to be the actual `https` URL, rather than the relative path to the docs.

PR #630, opened 2024-09-09
Fix K8s job status detection

pditommaso:
This PR fixes the detection of job status that can have one or more failed pods and still some active pods.

### DIFF
--- a/src/main/groovy/io/seqera/wave/service/k8s/K8sServiceImpl.groovy
+++ b/src/main/groovy/io/seqera/wave/service/k8s/K8sServiceImpl.groovy
@@ -205,7 +205,7 @@ class K8sServiceImpl implements K8sService {
                 .readNamespacedJob(name, namespace, null)
         if( !job )
             return null
-        if( job.status.active || !job.status.completionTime )
+        if( job.status.active )
             return JobStatus.Pending
         if( job.status.succeeded )
             return JobStatus.Succeeded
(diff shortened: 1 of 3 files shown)


munishchouhan:
While testing I found the issue, where if the job is failed wave stuck in checking the status because job.status.completionTime is null and this if remains true:
https://github.com/seqeralabs/wave/blob/007256216ba4f1ed0210786c1ba4261915fd018f/src/main/groovy/io/seqera/wave/service/k8s/K8sServiceImpl.groovy#L208

pditommaso:
Are you saying the the job can complete having  `completionTime == null` ?

munishchouhan:
> Are you saying the the job can complete having `completionTime == null` ?

when job failed, completionTime is still null, which makes this if statement true

munishchouhan:
ok i think we should remove `|| !job.status.completionTime` from here
because completionTime will only have value in case of success 
https://github.com/seqeralabs/wave/blob/007256216ba4f1ed0210786c1ba4261915fd018f/src/main/groovy/io/seqera/wave/service/k8s/K8sServiceImpl.groovy#L208

src/main/groovy/io/seqera/wave/service/k8s/K8sServiceImpl.groovy
```diff
@@ -205,11 +205,14 @@ class K8sServiceImpl implements K8sService {
                 .readNamespacedJob(name, namespace, null)
         if( !job )
             return null
-        if( job.status.succeeded==1 )
+        if( job.status.active || !job.status.completionTime )
```
munishchouhan, line 208:
```suggestion
        if( job.status.active )
```

pditommaso:
Gotcha!

Issue #559, opened 2024-07-09
Error "Request Entity Too Large"


pditommaso:
Submitting a build request using a large Conda environment, the following error is thrown: 

```
io.micronaut.http.client.exceptions.HttpClientResponseException: Client 'http://surrealv1:8100': Request Entity Too Large
	at io.micronaut.http.client.netty.DefaultHttpClient.handleStreamHttpError(DefaultHttpClient.java:1531)
	at io.micronaut.http.client.netty.DefaultHttpClient.lambda$streamRequestThroughChannel$47(DefaultHttpClient.java:1522)
	at reactor.core.publisher.FluxFlatMap$FlatMapMain.onNext(FluxFlatMap.java:386)
	at io.micronaut.reactive.reactor.instrument.ReactorSubscriber.onNext(ReactorSubscriber.java:57)
	at reactor.core.publisher.FluxCreate$BufferAsyncSink.drain(FluxCreate.java:814)
	at reactor.core.publisher.FluxCreate$BufferAsyncSink.next(FluxCreate.java:739)
	at reactor.core.publisher.FluxCreate$SerializedFluxSink.next(FluxCreate.java:161)
	at io.micronaut.http.client.netty.DefaultHttpClient.lambda$streamRequestThroughChannel0$48(DefaultHttpClient.java:1566)
	at io.netty.util.concurrent.DefaultPromise.notifyListener0(DefaultPromise.java:590)
	at io.netty.util.concurrent.DefaultPromise.notifyListenersNow(DefaultPromise.java:557)
	at io.netty.util.concurrent.DefaultPromise.notifyListeners(DefaultPromise.java:492)
	at io.netty.util.concurrent.DefaultPromise.setValue0(DefaultPromise.java:636)
	at io.netty.util.concurrent.DefaultPromise.setSuccess0(DefaultPromise.java:625)
	at io.netty.util.concurrent.DefaultPromise.trySuccess(DefaultPromise.java:105)
	at io.micronaut.http.client.netty.DefaultHttpClient$StreamStreamHttpResponseHandler.buildResponse(DefaultHttpClient.java:2343)
	at io.micronaut.http.client.netty.DefaultHttpClient$StreamStreamHttpResponseHandler.buildResponse(DefaultHttpClient.java:2331)
	at io.micronaut.http.client.netty.DefaultHttpClient$BaseHttpResponseHandler.channelReadInstrumented(DefaultHttpClient.java:2092)
[...]

PR #642, opened 2024-09-12
Add support for conda lock file

munishchouhan:
depends upon https://github.com/seqeralabs/libseqera/pull/25
the above PR needs 'git revert d088604' before merging

1. upload lockfile to  bucket
2. download lockfile from bucket
3. link in build page


pditommaso:
Good start. To tell the truth, still not sure we should go ahead with this approach or just store the lock file in the surreal db like we are doing for the conda env, even tho it there's the possibility to have the same problem as https://github.com/seqeralabs/wave/issues/559

munishchouhan:
> Good start. To tell the truth, still not sure we should go ahead with this approach or just store the lock file in the surreal db like we are doing for the conda env, even tho it there's the possibility to have the same problem as #559

https://github.com/seqeralabs/wave/issues/559 will be solved in surrealdb version 2.0.0

pditommaso:
I'm bit confused by this comment in the [issue](https://github.com/surrealdb/surrealdb/pull/4670)

> SURREAL_HTTP_MAX_ML_BODY_SIZE (defaults to 4 GiB)
SURREAL_HTTP_MAX_SQL_BODY_SIZE (defaults to 1 MiB)
SURREAL_HTTP_MAX_RPC_BODY_SIZE (defaults to 4 MiB)
SURREAL_HTTP_MAX_KEY_BODY_SIZE (defaults to 16 KiB)
SURREAL_HTTP_MAX_SIGNUP_BODY_SIZE (defaults to 1 KiB)
SURREAL_HTTP_MAX_SIGNIN_BODY_SIZE (defaults to 1 KiB)
SURREAL_HTTP_MAX_IMPORT_BODY_SIZE  (defaults to 4 GiB)


It seems suggesting the default sql body size is 1 MB, instead the error we are hitting is much smaller.

munishchouhan:
> It seems suggesting the default sql body size is 1 MB, instead the error we are hitting is much smaller.

we are using /key routes, which is constrained by 16 KiB
`SURREAL_HTTP_MAX_KEY_BODY_SIZE (defaults to 16 KiB)`

munishchouhan:
if we use sql to store it then we can bypass this limit

pditommaso:
This sounds like a plan. please give it a try

munishchouhan:
> This sounds like a plan. please give it a try

ok sure

munishchouhan:
in latter case of generating conda lockfile from conda file, we still need another job to achieve that

ewels:
Can we print the lock file to stdout and then capture that from the build?

munishchouhan:
> Can we print the lock file to stdout and then capture that from the build?
we can do this:

```
FROM {{base_image}}
COPY --chown=$MAMBA_USER:$MAMBA_USER conda.yml /tmp/conda.yml
RUN micromamba install -y -n base -f /tmp/conda.yml \
    {{base_packages}}
    && micromamba env export --explicit > environment.lock \
    && cat environment.lock
    && micromamba clean -a -y
RUN 
USER root
ENV PATH="$MAMBA_ROOT_PREFIX/bin:$PATH"
```

I ran it for conda package 'bwa'
i got this in the stdout

```
#10 12.24 # This file may be used to create an environment using:
#10 12.24 # $ conda create --name <env> --file <this file>
#10 12.24 # platform: linux-aarch64
#10 12.24 @EXPLICIT
```

ewels:
Exactly - that works!

Were there a load of lines after the `@EXPLICIT` yeah?


We'd need to remove the line prefixes but that's all I think..

pditommaso:
A better approach (maybe) could be: 1) creating the container "locally"; 2) copy the lock file from the built container via buildkit; 3) uploading it to the registry. 

Something similar is done for singularity, [here](https://github.com/seqeralabs/wave/blob/9acd7cdb45127137ff92fbc14e577a012f34d6b1/src/test/groovy/io/seqera/wave/service/builder/DockerBuildStrategyTest.groovy#L179-L179).

munishchouhan:
> A better approach (maybe) could be: 1) creating the container "locally"; 2) copy the lock file from the built container via buildkit; 3) uploading it to the registry.
> 
> Something similar is done for singularity, [here](https://github.com/seqeralabs/wave/blob/9acd7cdb45127137ff92fbc14e577a012f34d6b1/src/test/groovy/io/seqera/wave/service/builder/DockerBuildStrategyTest.groovy#L179-L179).

Ok sure, I will try this one

src/main/groovy/io/seqera/wave/service/persistence/impl/SurrealPersistenceService.groovy
```diff
@@ -234,4 +238,34 @@ class SurrealPersistenceService implements PersistenceService {
         return result
     }
 
+    @Override
+    void saveCondaLock(String buildId, String condaLock) {
+        final query = """\
+                                INSERT into wave_conda_lock {
+                                    buildId: '$buildId',
+                                    condaLock = '$condaLock'
```
pditommaso, line 246: what if the lock file contains a `'` ?

munishchouhan: good point
I have not tested yet with surrealDB
I will use bytes datatype to save it

munishchouhan:
Tested locally:
Screenshot 2024-09-18 at 17 41 10

logs:
Screenshot 2024-09-18 at 17 28 52

munishchouhan:
This PR can only be tested on dev once we release new wave-utils after merging this PR
https://github.com/seqeralabs/libseqera/pull/25

src/main/resources/io/seqera/wave/build-view.hbs
```diff
@@ -117,6 +117,13 @@
                 <td><a href="{{scan_url}}">Image scan report</a></td>
             </tr>
         {{/if}}
+
+        {{#if build_success}}
+            {{#if build_conda_lock_url }}
+                <td>Conda Lockfile</td>
+                <td><a href="{{build_conda_lock_url}}">Conda Lockfile</a></td>
```
ewels, line 124: `lockfile` is lower case. I'd actually prefer it as two words, but the [official conda docs](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#create-explicit-lockfiles-without-creating-an-environment) call it that, so let's stick to it for consistenct.

```suggestion
                <td>Conda lockfile</td>
                <td><a href="{{build_conda_lock_url}}">Conda lockfile</a></td>
```

pditommaso:
Do we have an estimation how big can be a lock file? If it's not too big it could be stored in the db as the conda env file.

munishchouhan:
not able to find anything on internet,

this is from chatgpt:
```
1. Conda Lock Files (conda-lock.yml)
Size Range: A typical Conda lock file can range from a few kilobytes to several megabytes, depending on the number of dependencies and platforms included.
Small environment (few dependencies): Can be under 100 KB.
Larger environments (many dependencies, or multi-platform): Can reach several MB (1-5 MB or more).

Real-World Examples:
Conda lock files for a moderately complex data science project (with dependencies like numpy, pandas, tensorflow, etc.) can easily be between 500 KB to 2 MB.
A web application with many dependencies across platforms (npm, pipenv, etc.) might generate a lock file of 2-5 MB.
```

pditommaso:
Umm, seems to big. @pinin4fjords do you any clue how big can be - on average - a conda lock file?

ewels:
* `multiqc`: 13K
* `bwa` + `samtools`: 3.4K
* `fastqc`: 8.2K
* `numpy` + `pandas`: 4.8K
* `cuda`: 20K

Struggled to get any huge lists of packages to build, so instead pulled out the largest `environment.yml` files in nf-core/modules:

* [`spotiflow`](https://github.com/nf-core/modules/blob/master/modules/nf-core/spotiflow/environment.yml): build failed
* [`normaldb`](https://github.com/nf-core/modules/blob/master/modules/nf-core/purecn/normaldb/environment.yml): 28K
* [`intervalfile`](https://github.com/nf-core/modules/blob/master/modules/nf-core/purecn/intervalfile/environment.yml): 28K
* [`imc2mc`](https://github.com/nf-core/modules/blob/master/modules/nf-core/mcstaging/imc2mc/environment.yml): 15K

They're still not that big, so I went back in time to some DSL1 pipelines, where we had a single conda environment for the entire pipeline:

* [rnaseq 1.4.2](https://github.com/nf-core/rnaseq/blob/1.4.2/environment.yml): _waiting, will report back when done_
* [Sarek 2.7](https://github.com/nf-core/sarek/blob/2.7.2/environment.yml): _waiting, will report back when done_
* [methylseq 1.6.1](https://github.com/nf-core/methylseq/blob/1.6.1/environment.yml): _waiting, will report back when done_

ewels:
So even with the biggest real-life conda environment that I could find, we're no-where near to the MB range..

pditommaso:
This is a good point to save the lock directly in the db. it would make it much simpler

pinin4fjords:
What is the upper limit this would place on the size of the locks? I think we all know that even if locks from nf-core are small, there will be that customer that does something funky to get us into the MB range.

munishchouhan:
> What is the upper limit this would place on the size of the locks? I think we all know that even if locks from nf-core are small, there will be that customer that does something funky to get us into the MB range.

1MB

pinin4fjords:
> > What is the upper limit this would place on the size of the locks? I think we all know that even if locks from nf-core are small, there will be that customer that does something funky to get us into the MB range.
> 
> 1MB

OK, then I'm going to out on a limb and suggest that this would eventually come back to bite us.

src/main/groovy/io/seqera/wave/service/logs/BuildLogServiceImpl.groovy
```diff
@@ -97,6 +100,7 @@ class BuildLogServiceImpl implements BuildLogService {
     @Override
     void storeLog(String buildId, String content){
         try {
+            content = removeCondaLockFile(content)
```
pditommaso, line 103: Thinking more about this, a better approach could be to extract the conda lock from the log and upload it in the same bucket just using a diffent key/extension just after the log. This would avoid having a "conda lock" service and extra logic to manage the surreal table. What do you think?

munishchouhan: this make sense, because conda lock depends upon log.

But it will make condalock coupled with log service, which make code less maintainable.
and if a customer wants to use different bucket for condalock, then this approach wont work.

we can have separate service for condalock and separate s3client and we can use the same bucket as logs in that client too. in this way it will be provide a way to use different bucket if required 
and also changes to log service wont affect condalock functionality

pditommaso: Not that bad, the lesser option the better, and ultimately the lock cannot exist without the log file.

munishchouhan: understood, I will make the changes

pditommaso:
@munishchouhan any progress on this?

munishchouhan:
> @munishchouhan any progress on this?

i will complete it today

src/main/groovy/io/seqera/wave/service/logs/BuildLogServiceImpl.groovy
```diff
@@ -69,34 +73,41 @@ class BuildLogServiceImpl implements BuildLogService {
...
-        final base = org.apache.commons.lang3.StringUtils.strip(prefix, '/')
+        final base = strip(prefix, '/')
         return "${base}/${buildId}.log"
     }
 
     @EventListener
     void onBuildEvent(BuildEvent event) {
         if(event.result.logs) {
             CompletableFuture.supplyAsync(() -> storeLog(event.result.id, event.result.logs), ioExecutor)
+            CompletableFuture.supplyAsync(() -> storeCondaLock(event.result.id, event.result.logs), ioExecutor)
```
pditommaso, line 102: I think there's no need to expose `storeCondaLock` independently, the storing of the lock file can stay in  `storeLog` implementation since it's a byproduct of the logs

munishchouhan: done

munishchouhan:
tested locally:
1. view
Screenshot 2024-10-03 at 14 51 18

2. condalock file
Screenshot 2024-10-03 at 14 51 31

3. s3 bucket
Screenshot 2024-10-03 at 14 51 10

src/main/groovy/io/seqera/wave/controller/BuildController.groovy
```diff
@@ -85,6 +86,18 @@ class BuildController {
...
 
+    @Produces(MediaType.TEXT_PLAIN)
+    @Get(value="/v1alpha1/builds/{buildId}/condalock")
+    HttpResponse<StreamedFile> getCondaLock(String buildId){
+        if( logService==null )
+            throw new IllegalStateException("Build Logs service not configured")
+        final condaLock = logService.fetchCondaLock(buildId)
+        return condaLock
+                ? HttpResponse.ok(condaLock)
+                    .header(HttpHeaders.CONTENT_DISPOSITION, "attachment; filename=\"" + buildId  + ".lock\"")
```
pditommaso, line 97:
```suggestion
                    .header(HttpHeaders.CONTENT_DISPOSITION, "attachment; filename=\"conda-env-" + buildId  + ".lock\"")
```

src/main/groovy/io/seqera/wave/service/logs/Conda.groovy
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Interface for Conda constants
+ *
+ * @author Munish Chouhan <munish.chouhan@seqera.io>
+ */
+interface Conda {
+
+    String CONDA_LOCK_START = "conda_lock_start"
+    String CONDA_LOCK_END = "conda_lock_end"
```
pditommaso, line 11: Let's keep in the log service

src/main/resources/io/seqera/wave/build-view.hbs
```diff
@@ -117,6 +117,13 @@
                 <td><a href="{{scan_url}}">Image scan report</a></td>
             </tr>
         {{/if}}
+
+        {{#if build_success}}
+            {{#if build_conda_lock_url }}
+                <td>Conda lockfile</td>
+                <td><a href="{{build_conda_lock_url}}">Conda lockfile</a></td>
+            {{/if}}
```
pditommaso, line 125: let's render the conda lock text after the conda environment block

pditommaso (changes requested):
Nice progress, just made a few comments

src/main/resources/io/seqera/wave/build-view.hbs
munishchouhan: do you mean after conda file?
https://github.com/seqeralabs/wave/blob/11816def0fea5cd4c0a24a6255a027e1752a0765/src/main/resources/io/seqera/wave/build-view.hbs#L132-L136

pditommaso: Yes, using a separate code block to show the lock file

munishchouhan: should we show the conda lockfile text in build page or just download link?

munishchouhan: ok same like logs, where we truncate after max size

should i put the same max size for condalock as logs?

munishchouhan: so only test block and no download link?

pditommaso: yep

munishchouhan: done
Screenshot 2024-10-03 at 16 05 04

pditommaso: Nice! any chance to add a copy & past action like in github code blocks 

Screenshot 2024-10-03 at 10 23 27

src/main/groovy/io/seqera/wave/service/logs/BuildLogService.groovy
```diff
@@ -28,6 +28,9 @@ import io.micronaut.http.server.types.files.StreamedFile
  */
 interface BuildLogService {
 
+    String CONDA_LOCK_START = "conda_lock_start"
```
pditommaso, line 31:
```suggestion
    String CONDA_LOCK_START = "<<<<<<< CONDA_LOCK_START"
```

src/main/groovy/io/seqera/wave/service/logs/BuildLogService.groovy
```diff
@@ -28,6 +28,9 @@ import io.micronaut.http.server.types.files.StreamedFile
...
+    String CONDA_LOCK_END = "conda_lock_end"
```
pditommaso, line 32:
```suggestion
    String CONDA_LOCK_END = ">>>>>>> CONDA_LOCK_END"
```

Git marker style

pditommaso: also it would be better to move in the implementation with `private static final`

src/main/groovy/io/seqera/wave/service/logs/BuildLogService.groovy
```diff
@@ -28,6 +28,9 @@ import io.micronaut.http.server.types.files.StreamedFile
  */
 interface BuildLogService {
 
+    String CONDA_LOCK_START = ">>>>>>> CONDA_LOCK_START"
```
pditommaso, line 31:
```suggestion
    String CONDA_LOCK_START = "<<<<<<< CONDA_LOCK_START"
```

src/main/groovy/io/seqera/wave/service/logs/BuildLogServiceImpl.groovy
```diff
@@ -69,21 +70,25 @@ class BuildLogServiceImpl implements BuildLogService {
     @Value('${wave.build.logs.maxLength:100000}')
     private long maxLength
 
+    @Nullable
+    @Value('${wave.build.logs.conda-lock-prefix}')
```
pditommaso, line 74:
```suggestion
    @Value('${wave.build.conda-lock-prefix}')
```

true that's the logs service, but at config level can be confusing

src/main/groovy/io/seqera/wave/service/logs/BuildLogService.groovy
munishchouhan: done

munishchouhan:
changes also made in libseqera
https://github.com/seqeralabs/libseqera/pull/25

src/main/groovy/io/seqera/wave/service/logs/BuildLogService.groovy
munishchouhan: done

munishchouhan:
this is how it looks now:
Screenshot 2024-10-03 at 16 51 00

pditommaso:
Good, what about this https://github.com/seqeralabs/wave/pull/642#discussion_r1786317757

src/main/resources/io/seqera/wave/build-view.hbs
munishchouhan: sorry missed it, will check it out

pditommaso:
I guess chatgpt can write the code for you for this 😄

munishchouhan:
@pditommaso is this ok?

https://github.com/user-attachments/assets/c8a92042-f3a7-489a-9ca6-5ce88c015fd3

pditommaso:
looks great! can can it be applied to all code blocks?

pditommaso:
Let's move ahead this pr \*without* copy & past and action. We'll add it consistently to all text blocks in a separate PR

munishchouhan:
> Let's move ahead this pr *without* copy & past and action. We'll add it consistently to all text blocks in a separate PR

ok sure, i will create another PR for that

munishchouhan:
> What happens if the request specifies a lock file URI instead of some packages?
> 
> https://github.com/seqeralabs/wave/blob/a59e423df6111e86ebc2a9882abac0fc6a90051e/src/main/groovy/io/seqera/wave/util/ContainerHelper.groovy#L96-L116
> 
> Does it work correctly?

we are creating lockfile in both cases, when we have a condafile and when we have only packages
https://github.com/seqeralabs/libseqera/pull/25/files

and then using micromamba to create lock file, so if the request is valid micromamba will create lock file

pditommaso:
ok. was it tested with singularity?

munishchouhan:
> ok. was it tested with singularity?

I was not aware about its support in singularity too

I will add it in singularity files and test it

munishchouhan:
changes to containerfiles added here:
https://github.com/seqeralabs/libseqera/pull/25

pditommaso:
It could be wrapped in a scrolling area

munishchouhan:
> It could be wrapped in a scrolling area

i will add a scroll

munishchouhan:
scroll bar added 

https://github.com/user-attachments/assets/069e3bdc-560a-411b-8581-14e6ba20e7cf

pditommaso:
Added `wave-utils@0.14.0` and shortened the lock markers. Tried a build but cannot see any lock file. @munishchouhan what am i missing?

munishchouhan:
> Added `wave-utils@0.14.0` and shortened the lock markers. Tried a build but cannot see any lock file. @munishchouhan what am i missing?

may be`wave.build.conda-lock-prefix`

pditommaso:
> may bewave.build.conda-lock-prefix

That was missing indeed, now I see the upload but it's not rendering in the view. checking

pditommaso:
ok works

munishchouhan:
@pditommaso is this integration test is enough or were you referring to some other test?
https://github.com/seqeralabs/wave/blob/fbe5f99983e2ae98671af579ea7a6e97216dfdd0/src/test/groovy/io/seqera/wave/service/logs/BuildLogsServiceTest.groovy#L37

pditommaso:
Ideally it would nice to have a test passing trough the controller, building a conda container and check the lock file exists. S3 uploading not so important, could be mocked. More important lock file is grabbed.

munishchouhan:
> Ideally it would nice to have a test passing trough the controller, building a conda container and check the lock file exists. S3 uploading not so important, could be mocked. More important lock file is grabbed.

I gave it try and able to succeed with full flow, what when we do build in test and image is uploaded to the registry, next time when the same test runs, it will find that image and return cached and in that cache we will not have conda lockfile 

I will try to mock the build and send mockedevent to do this testing

munishchouhan:
I overcome the caching issue by providing the invalid repo. so the push image part fails and everytime we run the test, it creates a fresh build

munishchouhan:
this e2e test tasks long time to complete, what we can do is add another module for e2e test, which runs on demand and when merging a pr to master

munishchouhan:
> Ok, let's keep the e2e in a separate PR and merge this one

yes, working on that, 
Ok I will remove the e2e from this PR

pditommaso:
this still requires the update in `libseqera` right?

munishchouhan:
Tested on local:
Screenshot 2024-10-07 at 15 17 29

PR #645, opened 2024-09-14
Improve job dispatcher

pditommaso:
This PR improves the Job dispatcher logic: 
1. use a common logic for loading the state record 
2. remove the JobEvent class 
3. automatic detection on Job handlers


pditommaso:
- use a common logic for loading the state record
- remove the JobEvent class
- automatic detection on Job handlers

PR #649, opened 2024-09-18
Wave inspect: CSS edits

ewels:
Minor improvements in CSS and styling for PR #619

* Encode SVG into the file, rather than including a new external asset
* Use [Inter](https://fonts.google.com/specimen/Inter) for our main font
* Use monospace fonts for technical details like name, tag, digest etc.
* Use [Roboto Mono](https://fonts.google.com/specimen/Roboto+Mono) for monospace font
* Avoid using inline CSS styling
* Set maximum width for first table column to avoid bounce when expanding tree nodes
* Minor tidying of padding and spacing

Future stuff we should do across all reports, for future PR(s):

- [ ] Do all of the above tweaks in all reports
- [ ] base64 encode the PNG images and include inline too, rather than using external image files
- [ ] Update table styling
- [ ] Include link to Seqera Wave webpage, in addition to the hostname of where it was generated
- [ ] General design updates to match other design assets (heading styles etc etc)

Wanted to get this in first though for PR #619, before tackling that larger effort across all reports.

Currently looks almost identical, which was the intention. Just minor tidying here:

CleanShot 2024-09-18 at 09 15 14@2x

### DIFF
--- a/src/main/resources/io/seqera/wave/assets/expand-collapse.svg
+++ b/src/main/resources/io/seqera/wave/assets/expand-collapse.svg
@@ -1,7 +1,0 @@
-<?xml version="1.0"?>
-<svg xmlns="http://www.w3.org/2000/svg" width="24" height="12" viewBox="0 0 40 20">
-  <g fill="#fff">
-    <path d="m5 9h4v-4h2v4h4v2h-4v4h-2v-4h-4z"/>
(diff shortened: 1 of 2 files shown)


munishchouhan (approved):

munishchouhan:
issue created for future stuff`
https://github.com/seqeralabs/wave/issues/650

PR #683, opened 2024-10-10
remove monospace from <td>

munishchouhan:
this PR will remove monospace from most of the <td>


munishchouhan:
> Let's remove from any table cells. We redesign this page later on

done

PR #692, opened 2024-10-15
Propagate container platform with freeze

pditommaso:
This PR fixes the propagation of requested container platform when determining the container entry point in a Fusion build.


src/main/groovy/io/seqera/wave/core/ContainerAugmenter.groovy
```diff
@@ -529,7 +534,7 @@ class ContainerAugmenter {
         }
 
         final manifestResult
-                = parseManifest(type, manifestsList,digest)
+                = parseManifest(type, manifestsList, digest)
```
pditommaso, line 537:
```suggestion
                = parseManifest(type, manifestsList,digest)
```

PR #697, opened 2024-10-16
fixed empty array plus icon

munishchouhan:
This PR will change the plus array for an empty to disable icon


munishchouhan:
tested locally:
Screenshot 2024-10-16 at 12 50 01

pditommaso:
Nice

PR #706, opened 2024-10-21
fixed inspect view

munishchouhan:
Annotations is a Map, so i added a length check for object type too, previously it was added only for arrays


munishchouhan:
Tested:
Screenshot 2024-10-21 at 09 07 10

PR #707, opened 2024-10-21
Render html from openapi spec generated by typespec

munishchouhan:
This PR will do the followings:
1. Move typespec to resources
2. create gradle task to install and compile typespec to generate openapi.yaml
3. create html to render openapi using swagger ui


munishchouhan:
> Not sure we should include this stuff within wave backend

so we want to host them separately?

pditommaso:
Think would make more sense, either in the main docs or a mini dedicated app. Tagging @ewels

ewels:
Yup, I'd expect to keep just the typespec in the repo and have everything else compiled and hosted separately. Where that happens I don't have a strong opinion on - either the Wave service itself? Or the docs website? (NB: Not the docs _repo_).

Tagging @llewellyn-sl @andrew-dawson-seqera

ewels:
Looks like the Platform API docs are hosted at https://cloud.seqera.io/openapi/index.html

So Wave docs could live at https://wave.seqera.io/openapi/index.html for example?

pditommaso:
That's fine. Then we need to create a mini nginx service to render the docs.

munishchouhan:
Screenshot 2024-10-23 at 13 07 15

pditommaso (changes requested):
Nice, then we need to including this container build in the same process  for the wave app container.

munishchouhan:
> Nice, then we need to including this container build in the same process for the wave app container.

ok sure

ewels:
Not sure what the web UI is there - any chance that we can use Swagger so that it has the same look + feel as https://cloud.seqera.io/openapi/index.html please?

munishchouhan:
> Not sure what the web UI is there - any chance that we can use Swagger so that it has the same look + feel as https://cloud.seqera.io/openapi/index.html please?

ok I will look into it

munishchouhan:
now using swagger ui
Screenshot 2025-01-15 at 15 17 33

pditommaso:
Nice

pditommaso:
What's missing to make the live?

munishchouhan:
> What's missing to make the live?

I will do a test  release, once it's successful, then I will work on infra part

typespec/index.html
```diff
@@ -0,0 +1,28 @@
...
+    <link rel="stylesheet" href="https://unpkg.com/swagger-ui-dist/swagger-ui.css">
+</head>
+<body>
+<div id="swagger-ui"></div>
+<script src="https://unpkg.com/swagger-ui-dist/swagger-ui-bundle.js"></script>
+<script src="https://unpkg.com/swagger-ui-dist/swagger-ui-standalone-preset.js"></script>
+<script>
+    window.onload = function() {
+        const ui = SwaggerUIBundle({
+            url: "tsp-output/@typespec/openapi3/openapi.yaml",
```
gavinelder, line 16: Can we change this URL so it is looking for resources under a single path of `/openapi`

```suggestion
            url: "openapi/openapi.yaml",
```

typespec/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM nginx:alpine
+
+# Copy the index.html and openapi.yaml to the Nginx html directory
+COPY index.html /usr/share/nginx/html/
+COPY tsp-output/@typespec/openapi3/openapi.yaml /usr/share/nginx/html/tsp-output/@typespec/openapi3/
```
gavinelder, line 5:
```suggestion
COPY tsp-output/@typespec/openapi3/openapi.yaml /usr/share/nginx/html/openapi/openapi.yaml
```

typespec/index.html
munishchouhan: changing it to a custom path, makes it harder to test locally the API changes, so keeping it same as typespec output path servers in every env

gavinelder: This would require an additional ingress path to be created for the route `tsp-output` on the kubernetes deployment.

gavinelder: Looking at Platform API Spec.

The UI is at 
https://cloud.seqera.io/openapi

With the schema being hosted at 
https://cloud.seqera.io/openapi/seqera-api-latest.yml

I still feel like we should align to 

https://wave.seqera.io/openapi/

And 

https://cloud.seqera.io/openapi/somefile.yaml

instead of hosting it on `tsp-output` especially since this change only impacts the docker container and not local dev flows.

munishchouhan: in the case of the platform, it's different because Openapi resides inside the platform, and in the case of Wave, it's a different app. I tried these changes but still facing the routing issue

munishchouhan:
tested with new port and wave dev:
Screenshot 2025-01-20 at 14 01 47

typespec/routes.tsp
```diff
@@ -2,12 +2,19 @@ import "./models/models.tsp";
...
+using TypeSpec.Versioning;
 
+@versioned(wave.Versions)
 @service({
-  title: "Wave service",
+  title: "Wave service"
 })
-@server("https://wave.seqera.io", "wave endopint")
+@server("https://wave.dev-tower.net", "wave dev endpoint")
+@server("https://wave.stage-seqera.io", "wave stage endpoint")
```
gavinelder, line 12: I wonder if this could be the base URL

munishchouhan: its selected by default
Screenshot 2025-01-20 at 16 43 44

munishchouhan: please let me know, if you meant something else

munishchouhan:
@pditommaso it is ready for review
tested:
Screenshot 2025-01-20 at 19 01 13

.github/workflows/typespec.yml
```diff
@@ -30,11 +30,26 @@ jobs:
...
+          aws-access-key-id: ${{secrets.TOWER_CI_AWS_ACCESS}}
+          aws-secret-access-key: ${{secrets.TOWER_CI_AWS_SECRET}}
+          aws-region: eu-west-1
+
+      - name : Login to Amazon ECR
+        id : login-ecr
+        uses : aws-actions/amazon-ecr-login@v1
+
+      - name: Release docs
+        if: "contains(github.event.head_commit.message, '[release docs]')"
```
pditommaso, line 53:
```suggestion
      - name: Release OpenAPI docs
        if: "contains(github.event.head_commit.message, '[release]')"
```

pditommaso: better to keep this aligned with the release

pditommaso (changes requested):

.github/workflows/typespec.yml
munishchouhan: we need to make sure, change the version in route.tsp also at the time of release

typespec/routes.tsp
munishchouhan: I think you meant if we can use the URL that we used to access the swagger UI
but typespec currently does not support it
https://typespec.io/docs/libraries/http/reference/decorators/#@TypeSpec.Http.server

munishchouhan:
tested in dev

https://github.com/user-attachments/assets/96b27999-346b-4772-9711-5e36142015a5

src/main/groovy/io/seqera/wave/controller/OpenAPIController.groovy
```diff
@@ -0,0 +1,41 @@
...
+ */
+@CompileStatic
+@Controller
+@ExecuteOn(TaskExecutors.BLOCKING)
+class OpenAPIController {
+    @Get(uri = "/openapi")
+    HttpResponse getOpenAPI() {
+    println("Redirecting to /openapi/")
+        HttpResponse.redirect(URI.create("/openapi/"))
+    }
```
pditommaso, line 40: Let's move this here, that's done mostly for it 

https://github.com/seqeralabs/wave/blob/5d5fd91352744febfc010708e326467d24bef798/src/main/groovy/io/seqera/wave/controller/ServiceInfoController.groovy#L45-L60

typespec/tag-and-push-docs.sh
pditommaso, line 1: `typespec/tag-and-push-openapi.sh`

typespec/routes.tsp
```diff
@@ -2,12 +2,19 @@ import "./models/models.tsp";
...
+@server("https://wave.seqera.io", "wave prod endpoint")
 namespace wave {
+    enum Versions {
+      v1: "1.16.8",
```
pditommaso, line 16: should this be updated manually?

munishchouhan: yes for now, typespec.yml only runs when there is a change in typespec folder and to make a doc release, we need a change there in the commit

munishchouhan: yes, that can be an option, i will move doc release to build.yml too
to not rely on typespec chnages

munishchouhan: ok better addin  VERSION path in typespec.yml, so that it runs when we change VERSION

munishchouhan:
Tested successfully in dev
Screenshot 2025-01-26 at 21 45 45
Screenshot 2025-01-26 at 21 45 21

typespec/tag-and-push-docs.sh
pditommaso, line 1: `tag-and-push-openapi.sh`

munishchouhan: done

typespec/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM nginx:alpine
+
+# Copy index.html and openapi.yaml to the Nginx html directory
+COPY index.html /usr/share/nginx/html/openapi/index.html
+COPY tsp-output/@typespec/openapi3/openapi.yaml /usr/share/nginx/html/openapi/openapi.yaml
+
+# Use sed to modify the default Nginx configuration to listen on port 9080
+RUN sed -i 's/listen       80;/listen       9080;/' /etc/nginx/conf.d/default.conf
```
pditommaso, line 8: They provide an un unprivileged image that does not require this https://hub.docker.com/r/nginxinc/nginx-unprivileged

munishchouhan: done

pditommaso (approved):

PR #710, opened 2024-10-22
Build report: make inspect link separate to image name

ewels:
Give the inspect link its own text, rather than linking the container name (which itself is a URL, and behaves as such in emails giving different behaviour in different contexts).

CleanShot 2024-10-22 at 08 29 29@2x


pditommaso:
Why?

pditommaso:
Saw comment in slack

PR #724, opened 2024-10-28
fixed inspect view

munishchouhan:
This PR will fix the case, where its prints [object Object] for an empty map


munishchouhan:
tested:
empty map:
Screenshot 2024-10-28 at 16 21 47

non empty map:
Screenshot 2024-10-28 at 16 21 36

pditommaso (approved):

PR #728, opened 2024-10-31
Cap number of scan vulnerabilities reported 

munishchouhan:
This PR will add:
1. limit for storing vulnerabilities
2. ~~Change storing vulnerabilities to async~~

### DIFF
--- a/src/main/groovy/io/seqera/wave/service/persistence/impl/SurrealPersistenceService.groovy
+++ b/src/main/groovy/io/seqera/wave/service/persistence/impl/SurrealPersistenceService.groovy
@@ -28,7 +28,6 @@ import io.micronaut.core.annotation.Nullable
 import io.micronaut.http.client.exceptions.HttpClientResponseException
 import io.micronaut.runtime.event.ApplicationStartupEvent
 import io.micronaut.runtime.event.annotation.EventListener
-import io.seqera.wave.configuration.ScanConfig
 import io.seqera.wave.core.ContainerDigestPair
 import io.seqera.wave.service.builder.BuildRequest
 import io.seqera.wave.service.mirror.MirrorEntry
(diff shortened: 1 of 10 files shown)


src/main/groovy/io/seqera/wave/service/persistence/impl/SurrealPersistenceService.groovy
```diff
@@ -257,8 +261,23 @@ class SurrealPersistenceService implements PersistenceService {
         final vulnerabilities = scanRecord.vulnerabilities ?: List.<ScanVulnerability>of()
 
         // save all vulnerabilities
+        int count = 0
         for( ScanVulnerability it : vulnerabilities ) {
-            surrealDb.insertScanVulnerability(authorization, it)
+            if ( ++count > scanConfig.vulnerabilityLimit )
```
pditommaso, line 266: Fear too simple. it should be filtered by vulnerability levels ie. first the critical ones, then high, etc until reach the 50 cut off. 

It may be better doing it adding a `filter` method in this class 

https://github.com/seqeralabs/wave/blob/d1d1a6228bac8013632742bfa6654438f1f9e1d9/src/main/groovy/io/seqera/wave/service/scan/TrivyResultProcessor.groovy#L35-L37

munishchouhan: I thought vulnerabilities are already sorted, I will check and add a filter

pditommaso: I see, I was not aware of that, still better to keep the filtering in its own method

pditommaso:
Focusing this PR only on taking top most important vulnerabilities

PR #737, opened 2024-11-10
Migration to virtual trheads - phase 1

pditommaso:
This PR introduces the use of `BLOCKING` tasks executors for controllers, `PairingWebSocket` and `TowerConnector`. 

When running over Java 21 (and later) the blocking tasks executor use virtual thread pool to carry out the tasks execution.


VERSION
```diff
@@ -1 +1 @@
-1.13.11
+1.14.0-B4
```
```suggestion
1.13.11
```

VERSION
```diff
@@ -1 +1 @@
-1.14.1
+1.14.2-B2
```
```suggestion
1.14.1
```

PR #746, opened 2024-11-17
Migration to virtual threads - phase 1

pditommaso:
This PR introduces the use of virtual threads for controllers


VERSION
```diff
@@ -1 +1 @@
-1.14.2-B3
+1.14.2-B4
```
```suggestion
1.14.2-B3
```

PR #750, opened 2024-11-19
Check block existence with object operation

pditommaso:
This PR change the logic for checking the existence of blob cache by using S3 operation instead of HTTP head method. 

This is needed to prevent that the result of the HTTP HEAD check for a non-existing object retuning 404 (not found) is cached by the underlying caching rule as documented [here](https://developers.cloudflare.com/r2/reference/consistency/#caching). 

The of S3 API should bypass the caching layer and therefore preventing this issue.

### DIFF
--- a/src/main/groovy/io/seqera/wave/service/blob/impl/BlobCacheServiceImpl.groovy
+++ b/src/main/groovy/io/seqera/wave/service/blob/impl/BlobCacheServiceImpl.groovy
@@ -17,7 +17,6 @@
  */
 package io.seqera.wave.service.blob.impl
 
-import java.net.http.HttpClient
 
 import groovy.transform.CompileStatic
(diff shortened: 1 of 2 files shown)


pditommaso:
@munishchouhan can you give it a try at your convenience. Tagging @gavinelder for visibility.

munishchouhan:
There is a problem with my colima setup using other option to test it

munishchouhan:
Tested locally 
successfully with 10 parallel transfers:
Screenshot 2024-11-20 at 12 36 00

munishchouhan:
cache:
Screenshot 2024-11-20 at 12 37 26

munishchouhan:
tested blob exists:
Screenshot 2024-11-20 at 13 43 56

munishchouhan (approved):

Issue #702, opened 2024-10-17
[Feature] Add option to define toleration on wave-build jobs to allow scheduling on tainted nodes


gavinelder:
The following duplicates https://github.com/seqeralabs/wave/issues/291. 

## Background. 

In the current setup, wave build pods can utilize the nodeSelector functionality as defined in the Wave config map. This allows users to specify node selection criteria, ensuring that certain workloads run on specific nodes. For example, a user can configure the nodeSelector in the Wave config map as follows:

```yaml
      build:
          nodeSelector:
            # this node selector binds the build pods to a separate cluster node group
            linux/amd64: 'service=wave-build'
            linux/arm64: 'service=wave-build-arm64'
```

This configuration ensures that wave build pods are scheduled on nodes labeled with `service=wave-build`. While this approach assigns these pods to specific nodes, it does not prevent other pods from being scheduled on those nodes, which can lead to unrelated workloads being scheduled on node-groups intended for wave. 

## Problem 

We have seen instances of wave becoming a noisy neighbor impacting other pods scheduled on the same node as the wave-build pods resulting in degraded performance and in some instances node failure when wave-build was configured without shared storage.

## Feature Request 

By adding support for tolerations as part of the job configuration we can allow cluster operators to taint nodes with relevant `NoSchedule` annotation preventing other workloads being configured on the relevant wave-build nodes. 

```sh
kubectl taint nodes node-name service=wave-build:NoSchedule
```
This taint would prevent any pods without the appropriate toleration from being scheduled on the nodes reserved for wave build jobs.

The configuration for the Wave build jobs would then look like the following for example, allowing the pod to tolerate the node taint. 

```yaml
spec:
  tolerations:
    - key: "service"
      operator: "Equal"
      value: "wave-build"
      effect: "NoSchedule"
```

## Possible approach 

A potential solution would be to allow end users to provide a Kubernetes-compliant list of Tolerations that can be added to the build jobs via the Wave service configuration. These tolerations would then be validated by the Wave service and injected into the build job pods during their creation.

By allowing users to specify tolerations in a format that directly follows the Kubernetes specification, we reduce the need for Wave to enforce specific business logic around pod scheduling, while ensuring that users can define their own scheduling constraints and preferences.

An example config would look like the following.

```yaml
      build:
          nodeSelector:
            # this node selector binds the build pods to a separate cluster node group
            linux/amd64: 'service=wave-build'
            linux/arm64: 'service=wave-build-arm64'
          Tolerations:
[...]

PR #751, opened 2024-11-20
Add option to define toleration on k8s jobs

munishchouhan:
This PR will add config to setup toleration for build k8s jobs


munishchouhan:
@gavinelder can I get nodes in dev with taints for and and arm to test this feature?

pditommaso:
@gavinelder is node selector AND taint rule really needed? I'd prefer keeping only one of the two

pditommaso:
@gavinelder can you please have a look at previous comment at your convenience

bebosudo:
`nodeSelector` and taints/tolerations are two different mechanisms to instruct the k8s scheduler:
- `nodeSelector: mylabel=xyz` tells the scheduler to only use nodes with a `mylabel` label equal to `xyz`
  * this doesn't prevent _other_ pods from using the nodes labelled with `mylabel=xyz`
  * in a context where the cluster is shared between wave build pods and non-wave pods it means that non-wave pods could end up using the wave nodes
  * if the `nodeSelector` label used doesn't match a specific nodegroup, like `kubernetes.io/arch=amd64`, it would mean that wave pods may get scheduled on non-wave nodes
- on the other hand, when a taint is added to a node, only pods with a matching toleration will be able to run on that node
  * this means that you could set a taint like `service=wave-builds` on all nodegroups dedicated to wave builds, and if wave build pods have a matching tolerations they could run on those nodes, but no other pod would run on the wave-dedicated nodes
  * so taints/tolerations are a mechanism to prevent "contamination" between pods of different kinds
  * taints/tolerations can be used in conjunction with nodeSelector: in the taint example above, you could mark all wave build nodegroups with a generic `service=wave-build` taint, then use `nodeSelector` to direct the scheduler on more specific nodes, like amd/arm nodes, or low-priority vs premium wave build nodegroups, etc

So IMHO we should keep both mechanisms, hope it makes sense

pditommaso:
I know, but want to simplify this, so if taint should our need, the other should be removed

gavinelder:
I believe that wave should support both as they have unique use-cases however  reviewing this code I don't see why we need to implement it in this fashion with both an inclusion of an AMD & ARM toleration as they are the same. 

There are additional manifest extensions we should be aware of at the same time as implementing this feature such as `container.apparmor.security.beta.kubernetes.io` being deprecated and replaced with https://kubernetes.io/docs/tutorials/security/apparmor/  and `ttlSecondsAfterFinished` which will auto-delete kubernetes pods once they have completed allowing us to remove the cleanup logic from wave.

pditommaso:
Please include and example how the taint you are expected to be defined

pditommaso:
Thinking more about this, I believe that instead of implementing this into Wave we should look into having a mutating admission controller that applies the taint or ant other policy based on generic pod annotation. 

There are good solution for this, for example https://kyverno.io/ and surely more are available.

bebosudo:
A mutation admission controller was already considered when opening the issue: https://github.com/seqeralabs/wave/issues/702#issue-2594548783
That delegates maintenance of wave build pods to the final user, instead of us making wave use the native solution.

pditommaso:
Yes, that's the point. It looks like requirements vary depending the target installation and therefore we are not going to support this in Wave other then generic annotation .

pditommaso:
We discussed internally and agreed to implement a solution based on Kyverno (or similar approach)

PR #757, opened 2024-11-25
[Docs] Create separate API limits page

justinegeffen:


docs/api-limits.mdx
gavinelder, line 1: I wonder if it would be better to include this content in a section around fair usage.

This would hopefully avoid a search collision on API and give us a chance to explain the value it provides to the community.

justinegeffen: Good question. 

If there's there other content that we currently have, or anticipate creating, that could be used to build out a fair use section I could do that. 

Alternatively, we could rename this "Usage limits" which opens it up for additional content.

PR #759, opened 2024-11-26
Add blocking executor to async caches

pditommaso:
This PR adds the use of blocking (aka virtual threads) Caffeine caches


VERSION
```diff
@@ -1 +1 @@
-1.15.3
+1.15.4-B1
```
```suggestion
1.15.3
```

PR #761, opened 2024-11-27
Workflow to sync changelog with Seqera docs

vladsavelyev:
Automation to push a changelog entry into [seqeralbas](https://github.com/seqeralabs/docs)

That requires settings a PAT:

* Log into the GitHub account you want to push PRs from (with admin rights for both repos)
* Go to Settings > Developer Settings > Personal Access Tokens > Tokens (classic)
* Click Generate new token
* Select scope to push to seqeralabs/docs
* Store the PAT as `SEQERALABS_DOCS_PAT` secret in wave repo

### DIFF
--- a/.github/workflows/seqera_docs_changelog.yml
+++ b/.github/workflows/seqera_docs_changelog.yml
@@ -1,0 +1,61 @@
+name: Push changelog to Seqera Docs
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      release_name:
+        description: 'Release version (e.g. 1.0.0)'
+        required: true
+      release_body:
+        description: 'Release changelog content'
+        required: true
+
+jobs:
+  update-docs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+          
+      - name: Clone seqeralabs/docs
+        run: |
+          git clone https://github.com/seqeralabs/docs.git seqeralabs-docs
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Create changelog file
+        run: |
+          mkdir -p seqeralabs-docs/changelog/wave
+          cat << EOF > seqeralabs-docs/changelog/wave/${{ github.event.release.name || inputs.release_name }}.mdx
+          ---
+          title: Wave ${{ github.event.release.name || inputs.release_name }}
+          date: $(date +%Y-%m-%d)
+          tags: [wave]
+          ---
+
+          ${{ github.event.release.body || inputs.release_body }}
+          EOF
+
+      - uses: actions/create-github-app-token@v1
+        id: generate-token
+        with:
+          app-id: ${{ secrets.DOCS_BOT_APP_ID }}
+          private-key: ${{ secrets.DOCS_BOT_APP_PRIVATE_KEY }}
+          owner: seqeralabs
+          repositories: docs
+        
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v7
+        with:
+          token: ${{ steps.generate-token.outputs.token }}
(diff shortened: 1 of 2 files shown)


gavinelder:
Hi we should create a Github App for this purpose then save these as Org Secrets. 

please see https://github.com/peter-evans/create-pull-request/blob/25b6871a4ebe4c3585f47c7a687ac6fd0ec0e32d/docs/concepts-guidelines.md

```
      - uses: actions/create-github-app-token@v1
        id: generate-token
        with:
          app-id: ${{ secrets.APP_ID }}
          private-key: ${{ secrets.APP_PRIVATE_KEY }}
          owner: owner
          repositories: fork-of-repo

      - uses: actions/checkout@v4

      # Make changes to pull request here

      - name: Create Pull Request
        uses: peter-evans/create-pull-request@v7
        with:
          branch-token: ${{ steps.generate-token.outputs.token }}
          push-to-fork: owner/fork-of-repo
          maintainer-can-modify: false
          ```

**Comment by vladsavelyev** [comment]
Thanks @gavinelder, I updated the workflow to use the app. Could you please take a look to verify it looks good?

File: .github/workflows/seqera_docs_changelog.yml [hunk 1]
```diff
@@ -0,0 +1,60 @@
+name: Push changelog to Seqera Docs
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      release_name:
+        description: 'Release version (e.g. 1.0.0)'
```
gavinelder, line 8: Can we validate the release version matches an expected release version format. 


```
      - name: Validate inputs
        run: |
          RELEASE_NAME="${{ github.event.release.name || inputs.release_name }}"
          if [[ ! $RELEASE_NAME =~ ^v[1-9]\.[0-9]{1,2}\.[0-9]{1,2}$ ]]; then
            echo "Error: Invalid release name format. Expected version between v1.0.0 and v9.99.99"
            exit 1
          fi
```

I appreciate that this regex does not meet the entire semver specification however it will cover our explicit usecase.

Side note this thread is a brilliant example of why it is so basic.  https://gist.github.com/jhorsman/62eeea161a13b80e39f5249281e17c39

.github/workflows/seqera_docs_changelog.yml
```diff
@@ -0,0 +1,60 @@
...
+        with:
+          app-id: ${{ secrets.DOCS_BOT_APP_ID }}
+          private-key: ${{ secrets.DOCS_BOT_APP_PRIVATE_KEY }}
+          owner: seqeralabs
+          repositories: docs
+        
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v7
+        with:
+          branch-token: ${{ steps.generate-token.outputs.token }}
```
gavinelder, line 50:
```suggestion
          token: ${{ steps.generate-token.outputs.token }}
          branch-token: ${{ steps.generate-token.outputs.token }}
```

https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#branch-token

```
The action first creates a branch, and then creates a pull request for the branch. For some rare use cases it can be useful, or even necessary, to use different tokens for these operations. It is not advisable to use this input unless you know you need to.
```

This action by default uses the token in the context of the action which I don't believe will have permission to create a pull request against the upstream repository.

.github/workflows/seqera_docs_changelog.yml
ewels: Why do we need to validate it?

I'm not sure that we should to be honest. By the time this action runs, the release is already minted. I think that the changelog should match the release name, whatever format that took.

gavinelder (approved):

.github/workflows/seqera_docs_changelog.yml
gavinelder: This is a fair point this also need to go through human review at the docs side.

Issue #776, opened 2024-12-19
Improve caching of blob redirect responses


pditommaso:
When serving container requests Wave behave as a proxy in between the Docker client and the target container registry. 

This means that to serve requests it's needed to connect the target registry, and therefore the quality of the service may be affected by the reliability of the target registry. 

To improve the service throughput all redirect responses from the target registry can be cached by wave to avoid to avoid reaching the remote service once the response has been collected. 

This goal can be achieved caching the result of [handleRequest](https://github.com/seqeralabs/wave/blob/d5c086dac3a41b1a3a4260ae84f980e58b579ace/src/main/groovy/io/seqera/wave/core/RegistryProxyService.groovy#L144-L144), when: 
- the response status is a [redirect code](https://github.com/seqeralabs/wave/blob/d5c086dac3a41b1a3a4260ae84f980e58b579ace/src/main/groovy/io/seqera/wave/WaveDefault.groovy#L44-L44)
- or, when it's returned a 2xx..3xx http status and the [response body is available](https://github.com/seqeralabs/wave/blob/d5c086dac3a41b1a3a4260ae84f980e58b579ace/src/main/groovy/io/seqera/wave/core/RegistryProxyService.groovy#L188-L188)

PR #778, opened 2024-12-19
Implement caching for proxy responses

pditommaso:
This improves the Wave caching for HTTP responses in two ways: 

1. Add a tiered-cache for HTTP requests made by the proxy service to access container blobs and the response is either a redirection or successful status code
2. Improve efficiency in the cached locking scheme, so that a separate lock object is used for each unique key 

Solve https://github.com/seqeralabs/wave/issues/776


VERSION
```diff
@@ -1 +1 @@
-1.16.0
+1.16.1-B1
```
```suggestion
1.16.0
```

PR #781, opened 2024-12-25
Remove unneeded user async api

pditommaso:
This PR removes the use of async API for `getUserByAccessToken` method since the use of virtual threads make it useless.


munishchouhan:
Tested 10 parallel builds with tower-token and there are no relevant errors in logs:
Screenshot 2024-12-26 at 15 50 09

munishchouhan (approved):

PR #782, opened 2024-12-26
Progressive container expiration extension

pditommaso:
### Summary

Ephemeral containers provisioned by Wave are accessible for a fixed period of time (36h) from the moment they are requested. That fixed window is a blunt compromise: too short for long workflow runs, too long for short ones, and unrelated to whether the workflow still needs the container.

This PR ties the accessibility of an ephemeral container to the lifecycle of the Platform workflow that requested it:

- **Kept alive while the run is active** — the container remains accessible for the whole duration of a running workflow.
- **Revoked shortly after the run ends** — access ends within a short window (~15 min) after the workflow completes, whether it succeeded or failed.
- **Hard lifetime cap** — a container is never accessible more than 48h after the original request, regardless of run state.

Only containers requested by a Platform workflow run are affected. Builds, mirrors and other requests keep the existing behaviour.

### How it works

- A workflow-bound container token is granted a short access window instead of the fixed 36h.
- A background watcher periodically checks the workflow status on Platform and renews the token while the run is active. When the run finishes — or the 48h cap is reached — it stops renewing and the token lapses on its own.
- If Platform can't be reached, the token is not renewed and simply lapses (with retry slack), so access never outlives what can be confirmed.
- The container view shows the current expiration.

### Behaviour / guarantees

- Accessible for the full duration of a running workflow, up to the 48h cap.
- Revoked within ~15 min of the run completing (succeeded or failed).
- Fail-safe: access is never extended beyond what the workflow status confirms.

### Test plan

- `ContainerRequestServiceImplTest` — watcher renewal logic: renews while active; stops on completed/failed/cancelled; fail-closed when the status can't be resolved; stops when the request is gone; stops at the 48h cap.
- `ContainerRequestRangeTest` — scheduling store stores/retrieves entries by time range.
- `TowerClientTest` — workflow-describe endpoint URL construction.
- Manual: verify a container tied to a long-running workflow stays accessible for the run's duration and becomes inaccessible shortly after it completes; verify the 48h cap is enforced.


munishchouhan:
@pditommaso Can we move forward with this PR?
or is it blocked by some other changes?

munishchouhan:
### relevance:
**Without this feature:** Container tokens have a fixed TTL (36h on master). If a workflow runs longer than that, the token expires and the container becomes inaccessible — the     
  workflow fails or can't pull its container.
                                                                                                                                                                                   
**With this feature:** A background watcher checks if the workflow is still running on Tower. If it is, the token TTL gets extended automatically (up to a max of 2 days). Once the workflow finishes, the token expires normally.
                                                                                                                                                                                   
It's essentially a keep-alive mechanism — tokens stay valid as long as the workflow needs them, but don't linger forever.

munishchouhan:
Tested locally:
**config**

```
  tokens:
    cache:
      duration: '10s'
      check-interval: '5s'
      max-duration: '60s'
    watcher:
      interval: '5s'
      delay: '2s'
```

**workflow**
Screenshot 2026-04-08 at 16 11 49

**wave logs**
```
16:08:52.547 [scheduled-executor-thread-17] DEBUG i.s.w.s.r.ContainerRequestServiceImpl - Container request '07e49374f8ba' does not require refresh - deadline=2026-04-08T14:08:56.056349Z; expiration=2026-04-08T14:09:01.056349Z >> wt=
16:08:52.547 [scheduled-executor-thread-17] TRACE i.s.w.s.r.ContainerRequestServiceImpl - Scheduling container request io.seqera.wave.service.request.ContainerRequestRange$Entry(07e49374f8ba, 1FMdRb3swdoqTh, 2026-04-08T14:09:01.056349Z) - event ts=2026-04-08T14:08:57.547582Z >> wt=
16:08:57.549 [scheduled-executor-thread-16] INFO  i.s.w.s.r.ContainerRequestServiceImpl - Container request '07e49374f8ba' expiration is extended by: PT13.507134S; at: 2026-04-08T14:09:11.056349Z; (was: 2026-04-08T14:09:01.056349Z) >> wt=
16:08:57.550 [scheduled-executor-thread-16] TRACE i.s.w.s.r.ContainerRequestServiceImpl - Scheduling container request io.seqera.wave.service.request.ContainerRequestRange$Entry(07e49374f8ba, 1FMdRb3swdoqTh, 2026-04-08T14:09:11.056349Z) - event ts=2026-04-08T14:09:02.550394Z >> wt=
16:09:02.549 [scheduled-executor-thread-20] DEBUG i.s.w.s.r.ContainerRequestServiceImpl - Container request '07e49374f8ba' does not require refresh - deadline=2026-04-08T14:09:06.056349Z; expiration=2026-04-08T14:09:11.056349Z >> wt=
16:09:02.549 [scheduled-executor-thread-20] TRACE i.s.w.s.r.ContainerRequestServiceImpl - Scheduling container request io.seqera.wave.service.request.ContainerRequestRange$Entry(07e49374f8ba, 1FMdRb3swdoqTh, 2026-04-08T14:09:11.056349Z) - event ts=2026-04-08T14:09:07.549954Z >> wt=
16:09:07.545 [scheduled-executor-thread-19] INFO  i.s.w.s.r.ContainerRequestServiceImpl - Container request '07e49374f8ba' expiration is extended by: PT13.511018S; at: 2026-04-08T14:09:21.056349Z; (was: 2026-04-08T14:09:11.056349Z) >> wt=
16:09:07.546 [scheduled-executor-thread-19] TRACE i.s.w.s.r.ContainerRequestServiceImpl - Scheduling container request io.seqera.wave.service.request.ContainerRequestRange$Entry(07e49374f8ba, 1FMdRb3swdoqTh, 2026-04-08T14:09:21.056349Z) - event ts=2026-04-08T14:09:12.546107Z >> wt=
16:09:12.550 [scheduled-executor-thread-4] DEBUG i.s.w.s.r.ContainerRequestServiceImpl - Container request '07e49374f8ba' does not require refresh - deadline=2026-04-08T14:09:16.056349Z; expiration=2026-04-08T14:09:21.056349Z >> wt=
16:09:12.550 [scheduled-executor-thread-4] TRACE i.s.w.s.r.ContainerRequestServiceImpl - Scheduling container request io.seqera.wave.service.request.ContainerRequestRange$Entry(07e49374f8ba, 1FMdRb3swdoqTh, 2026-04-08T14:09:21.056349Z) - event ts=2026-04-08T14:09:17.550524Z >> wt=
16:09:17.550 [scheduled-executor-thread-5] INFO  i.s.w.s.r.ContainerRequestServiceImpl - Container request '07e49374f8ba' expiration is extended by: PT13.508901S; at: 2026-04-08T14:09:31.056349Z; (was: 2026-04-08T14:09:21.056349Z) >> wt=
16:09:17.550 [scheduled-executor-thread-5] TRACE i.s.w.s.r.ContainerRequestServiceImpl - Scheduling container request io.seqera.wave.service.request.ContainerRequestRange$Entry(07e49374f8ba, 1FMdRb3swdoqTh, 2026-04-08T14:09:31.056349Z) - event ts=2026-04-08T14:09:22.550888Z >> wt=
16:09:22.550 [scheduled-executor-thread-12] DEBUG i.s.w.s.r.ContainerRequestServiceImpl - Container request '07e49374f8ba' does not require refresh - deadline=2026-04-08T14:09:26.056349Z; expiration=2026-04-08T14:09:31.056349Z >> wt=
16:09:22.550 [scheduled-executor-thread-12] TRACE i.s.w.s.r.ContainerRequestServiceImpl - Scheduling container request io.seqera.wave.service.request.ContainerRequestRange$Entry(07e49374f8ba, 1FMdRb3swdoqTh, 2026-04-08T14:09:31.056349Z) - event ts=2026-04-08T14:09:27.550450Z >> wt=
16:09:27.552 [scheduled-executor-thread-7] INFO  i.s.w.s.r.ContainerRequestServiceImpl - Container request '07e49374f8ba' reached max allowed duration - expiration=2026-04-08T14:09:31.056349Z; new expiration=2026-04-08T14:09:41.056349Z; workflow=1FMdRb3swdoqTh >> wt=

```

cristianrcv (approved):

munishchouhan:
@pditommaso should we merge this pr?

pditommaso:
let me review

munishchouhan:
@pditommaso please review

jonmarti (approved):

src/main/groovy/io/seqera/wave/configuration/ContainerRequestConfig.groovy
```diff
@@ -0,0 +1,110 @@
...
+         */
+        @Value('${wave.tokens.cache.max-duration:2d}')
+        Duration maxDuration
+
+        /**
+         * This method returns the period of time between two consecutive check events.
+         * The interval determines how frequently a refresh operation is triggered.
+         * A shorter interval means more frequent checks, while a longer interval reduces checks frequency.
+         */
+        @Value('${wave.tokens.cache.check-interval:30h}')
```
pditommaso: 🔴 **The `30h` default makes the extension feature completely inert.**

In `check0`, the extension step computes `newExpire = entry.expiration + checkInterval.multipliedBy(2)` and bails when `Duration.between(creationTime, newExpire) > maxDuration`. With the shipped defaults (`duration=36h`, `check-interval=30h`, `max-duration=2d`):

```
newExpire − creationTime = 36h + (2 × 30h) = 96h  >  48h  →  always "reached max allowed duration", return
```

So **no token is ever extended** — the first (and only) check always exceeds the max. For even one extension to succeed you need `duration + 2·checkInterval ≤ maxDuration`, i.e. `checkInterval ≤ 6h`.

The `30h` also breaks the step-2 "does not require refresh" skip: `deadline = expiration − 30h = creation+6h`, while the check first fires at `creation+30h`, so `now < deadline` is never true. This looks like a typo for `30m`. No `application.yml` override exists, so this default governs production.

src/main/groovy/io/seqera/wave/service/persistence/WaveContainerRecord.groovy
```diff
@@ -48,134 +48,134 @@ class WaveContainerRecord {
      * Wave request id, this will be the token
      */
     @PostgresIgnore
-    final String id
+    String id
```
pditommaso: 🟡 **All ~24 fields dropped from `final` to mutable to set one field.**

The only mutation this PR needs is `requestRecord.expiration = newExpire` in `check0`. Making the entire record mutable throws away the immutability of `id`, `user`, `fingerprint`, the digests, etc. Prefer keeping the rest `final` and exposing only `expiration` as mutable — or add a `withExpiration(...)` copy, matching the `Entry.withExpiration` pattern introduced in `ContainerRequestRange` in this same PR.

src/main/groovy/io/seqera/wave/service/request/ContainerRequestServiceImpl.groovy
```diff
@@ -74,4 +88,131 @@ class ContainerRequestServiceImpl implements ContainerRequestService {
...
+
+    protected void watch() {
+        final now = Instant.now()
+        final keys = containerRequestRange.getEntriesUntil(now, config.watcher.count)
+        for( Entry it : keys ) {
+            try {
+                check0(it, now)
+            }
+            catch (InterruptedException e) {
+                Thread.interrupted()
```
pditommaso: 🟢 **`Thread.interrupted()` clears and discards the interrupt instead of preserving it.**

`Thread.interrupted()` is a static check-and-clear that returns a boolean and throws the interruption away. To preserve shutdown semantics this should restore the flag:

```groovy
catch (InterruptedException e) {
    Thread.currentThread().interrupt()
}
```

As written, a scheduler shutdown that interrupts the watcher mid-cycle is silently swallowed.

pditommaso (commented):
Reviewed the TTL-extension watcher. The mechanism is well-structured and the range store correctly consumes entries on read (`getRange(remove=true)`), so there's no re-processing loop. Three findings inline, most severe first:

1. 🔴 `check-interval: 30h` default makes the feature inert — `newExpire = expiration + 2·checkInterval` (96h from creation) always exceeds `max-duration` (2d), so no token is ever extended. Likely a typo for `30m`.
2. 🟡 `WaveContainerRecord` drops all ~24 fields from `final` to mutable just to set `expiration`.
3. 🟢 `Thread.interrupted()` in `watch()` discards the interrupt instead of restoring it.

Finding #1 is the blocker: with the default config the watcher runs every 10s but never actually extends anything.

Minor note (not blocking): `describeWorkflow` is cached for 10m but its `Workflow.status` changes over a run (RUNNING→SUCCEEDED), so status can be up to 10m stale — fine given the check cadence, worth keeping in mind.

src/main/groovy/io/seqera/wave/configuration/ContainerRequestConfig.groovy
pditommaso: Fixed — default changed to `30m` (commit *fix: correct ephemeral container check-interval default to 30m*).

src/main/groovy/io/seqera/wave/service/persistence/WaveContainerRecord.groovy
pditommaso: Fixed — all fields kept `final`; added a `withExpiration()` copy method instead of an in-place setter (commit *refactor: keep WaveContainerRecord immutable when extending expiration*).

src/main/groovy/io/seqera/wave/service/request/ContainerRequestServiceImpl.groovy
pditommaso: Fixed — now uses `Thread.currentThread().interrupt()` to restore the flag (commit *fix: preserve interrupt flag in container request watcher*).

pditommaso:
### End-to-end validation ✅

Ran a local test against a real Platform workflow (`pditommaso/nf-sleep`, ~8 min run) and confirmed the full token lifecycle from `wave.log`.

**Workflow `24R0olI4VorPdj` (`gloomy_booth`):** submitted `19:27:49` → started `19:28:13` → **completed `19:36:21` (SUCCEEDED)**.

**Container token `161ef2c93432`** — `wave.ngrok.app/wt/161ef2c93432/nextflow/bash:latest`:

| Time | Event |
|------|-------|
| 19:20:57 | Watcher started — `access-ttl=15m; refresh-interval=4m30s; max-duration=48h; watcher-interval=10s` |
| 19:28:15 | Token created; advertised expiration = request time + 48h (ceiling) |
| 19:28:16–20 | Container pulled through the proxy — access works while the run is active |
| 19:32:51 | Refresh #1 → workflow `RUNNING` → **renewed** 15 min (→ expires 19:47:50) |
| 19:37:31 | Refresh #2 → fresh status `SUCCEEDED` → **stop tracking** (no renewal) |
| 19:47:50 | Token lapses at its last granted expiry — **~11.5 min after completion** |

**After the token lapsed, the container is correctly rejected:**

```
$ docker pull wave.ngrok.app/wt/161ef2c93432/nextflow/bash:latest
Error response from daemon: failed to resolve reference
"wave.ngrok.app/wt/161ef2c93432/nextflow/bash:latest": not found
```

**Guarantees confirmed:**
- ✅ Accessible for the full duration of the running workflow.
- ✅ Revoked ~11.5 min after completion — within the 15 min `access-ttl` bound.
- ✅ Completion detected via a fresh (uncached) Platform status check at the next refresh tick.
- ✅ Client is advertised the stable 48h ceiling, not the rolling internal TTL.

gavinelder:
@pditommaso I really like the work here. I have a couple of concerns, mainly around the fail-closed nature of the refresh path. We're introducing a new failure domain: a customer's pipeline that would previously have run to completion
with all its run-outputs can now have its container revoked mid-run simply because Platform lost connectivity for longer than the ~15 min refresh window. That's a regression in resilience for the running workflow, even though nothing
was actually wrong with it.

I'm also looking at the impact on Platform. This adds recurring API calls to the Platform instance to describe the workflow on every refresh tick. Wave's own container-request lookups are redis-backed and cheap, but the describe calls
hit Platform (and its DB) and are made per container — so a run that pulls N containers generates N independent status lookups per tick. We should look to optimize this: coalesce describe calls by workflowId so N containers from one
run share a single status lookup per tick rather than N independent reads. That should meaningfully reduce the added load.

Concretely, the things that would make me comfortable rolling this out:

1. Make the failure mode a config-driven flag (fail-closed by default, fail-open opt-in). Fail-open would keep renewing when the status can't be confirmed — still bounded by the 48h max-duration cap — so a transient pairing/Platform
blip doesn't take down a running workflow. This lets us pick the trade-off per deployment while we roll out.
2. Add Prometheus metrics around revocation under refresh failure — e.g. counters for refresh attempts, refresh failures (unreachable/unknown status), and tokens revoked specifically because the status couldn't be confirmed. That
gives us the data to make a metrics-driven decision on the right default and window rather than guessing.
3. Coalesce refresh lookups by workflowId (as above) to cap the per-run load on Platform, and expose the describe-call volume in the metrics from (2) so we can watch the real cost as adoption grows.

With the config flag, the metrics, and the coalescing, we can roll this out safely and tune the defaults from evidence.

gavinelder:
Also it would be nice if this feature could be configurable so that those self hosting can opt out for high traffic envs

pditommaso:
Thanks so much for the thoughtful review @gavinelder 🙏 — these were all great calls. Here's what I've done on this branch:

**1. Opt-out for high-traffic / self-hosted** — added `wave.tokens.watcher.enabled` (default `true`). Flip it off and every container request keeps the fixed cache duration with the status watcher never scheduled, so there's zero added Platform load. One clean switch to disable the whole feature. 👍

**2. Coalesce describe calls by workflowId** — done! Though I landed on a slightly different mechanism than per-tick coalescing, and I think it works out nicer. The catch is that the tokens of one run don't check on the same tick — each re-arms independently, so they end up scattered across the refresh window and a per-tick group-by would rarely catch two together. So instead `describeWorkflow` now goes through the existing short-duration client cache keyed by `workflowId`: all N tokens of a run collapse onto **one** Platform lookup per window no matter when they fire. Bonus, it's cross-replica since the client cache is redis-backed (L2). Completion shows up at most the cache TTL late, which is immaterial — `access-ttl` (minutes) is the real bound on post-completion access, not this cache.

**3. Metrics** — added a Prometheus counter `wave.tokens.refresh` tagged by `result`: `renewed`, `completed`, `unresolved` (status couldn't be confirmed → exactly the fail-closed revocation signal you flagged), `error` (Platform unreachable), `expired` (48h cap), and `gone`. That gives us refresh volume, failure rate, and unconfirmed-revocation rate so we can tune the defaults from real data. Describe-call volume is the sum of all results except `gone`.

**On fail-open** — here I'd like to gently push back and hold off for now, if you're open to it.

The exposure turns out to be narrower than a single blip: I widened the window to `access-ttl=20m` with `refresh-interval=180s`, so a token gets **6 renewal attempts** before it can ever lapse. Any *one* of those recovering saves it — so we ride out any Platform hiccup up to ~20 minutes. For it to actually revoke a running workflow's token, Platform would have to be unreachable from Wave *continuously* for a full 20 minutes. And at that point we're in a genuine outage where Wave requests for that workflow are already failing on all the other Platform calls anyway (credentials, pairing, etc.) — the token lapse really isn't what breaks the run.

Paired with the opt-out above (anyone who wants none of this can just switch it off), I think fail-closed with a wide retry window is the right default, and fail-open mostly buys us a config knob plus a lingering-access edge case for a scenario that's already degraded. That said, I'm totally happy to add it later if the `unresolved`/`error` metrics show it's actually biting in practice — which is exactly what they're there to tell us. 🙂

gavinelder (approved):
Thank you for the logical commits it made reviewing the incremental improvements easier. 

I will pick up a task to define appropriate operational alerts and runbooks to guide our decision making if we ever have to revert the behaviour at runtime.

pditommaso:
### Follow-up fix + end-to-end validation on stage

**Fix (`de6c1bbbd`):** the watcher rolled the persisted container-request expiration forward on each renewal via `saveContainerRequestAsync`, which performs an **INSERT** — the row already exists from request creation, so every renewal failed with `duplicate key value violates unique constraint "wave_request_pkey"` and the displayed expiration stayed frozen at the initial value. Replaced with `updateContainerExpirationAsync`, backed by an in-place JSONB `UPDATE` (mirroring `updateContainerDigests`). Redis TTL renewal / access control were never affected — this was the persisted record + error-log spam.

**End-to-end validation (stage, `v1.36.0-B2`):** ran a real multi-container Platform workflow (`t7misoleSegG4`, nf-core-style, 20+ tokens) and confirmed the full lifecycle from `wave.log`:

- **Token creation** — workflow-bound tokens advertise the stable 48h ceiling to the client, e.g. `expiration=2026-07-17T10:20Z`.
- **Renewal while active** — every ~3 min per token: `Fetched workflow status … status=RUNNING` → `renewed for PT20M; expires at …`, rolling the expiry forward. A token created at 10:20 was still alive and renewing ~54 min later, proving keep-alive across many windows while the run stayed active.
- **Completion → revocation** — on the first check after the run finished, a fresh (uncached) status was observed and tracking stopped:
  ```
  11:14:51  Fetched workflow status … 'ba6a9e800d95' … status=SUCCEEDED
  11:14:51  Container request 'ba6a9e800d95' - workflow t7misoleSegG4 completed; stop tracking
  ```
  No further renewal → the token lapses within its remaining access-ttl.
- **Coalescing** — tokens of the same run share a single Platform status lookup per window (short-cache keyed by `workflowId`), cross-replica via the redis-backed client cache.
- **Cross-replica** — the same token was renewed by different pods on different ticks; the redis-backed range store distributes work correctly.
- **Persistence fix confirmed** — **zero** `duplicate key` errors on B2 across the whole run; on B1 those same renewals threw every 3 min. The persisted expiration now rolls forward in place as intended.
- **Clean logs** — no `ERROR`/`WARN`/exceptions in stage since the B2 rollout (verified by confirming the same query still surfaces the pre-B2 duplicate-key errors).

Metrics (`wave.tokens.refresh{result}`) and the opt-out flag (`wave.tokens.watcher.enabled`) are also live and behaving as expected.

PR #783, opened 2024-12-26
775 use tiered cache in place of caffeine cache

munishchouhan:
Add the following:
1. AwsEcrCache
2. RegistryLookupCache
3. ~RegistryAuthCache~
4. TieredCacheKey


munishchouhan:
Tested locally with Redis:
1. 10 parallel builds with tower token 
2. 10 parallel builds with tower token and freeze enabled

munishchouhan:
Tested Dev:
1. 10 parallel builds with tower token
2. 10 parallel builds with tower token and freeze-enabled

src/main/groovy/io/seqera/wave/store/cache/AbstractTieredCache.groovy
```diff
@@ -293,4 +318,8 @@ abstract class AbstractTieredCache<V extends MoshiExchange> implements TieredCac
         l1.invalidateAll()
     }
 
+    void invalidate(K key) {
+        l1.invalidate(key.stableHash())
+    }
```
pditommaso, line 323: This is a problem because it removes the value only from the caffeine cache but it remains in redis, and even worse in the in-memory cache in other nodes that may have loaded that value

munishchouhan: for now, i have removed registryauthcache till the time we done find solution to invalidate it from all app instances

VERSION
```diff
@@ -1 +1 @@
-1.17.2
+1.17.3-B0
```
```suggestion
1.17.2
```

PR #785, opened 2025-01-07
Use caffeine for tiered cache lock cache

pditommaso:
Tiered cache implementation uses a `WeakHashMap` to allocate a `Lock` object for the current cache key. 

https://github.com/seqeralabs/wave/blob/db70fd83c9b15c865e3836ea50eff300d9c2c365/src/main/groovy/io/seqera/wave/store/cache/AbstractTieredCache.groovy#L174-L174

However the `WeakHashMap` is not thread-safe (!). This can lead to a not-deterministic behaviour causing the `computeIfAbsent` to hang while trying to acquire the lock object. 

This PR replaces the `WeakHashMap` with a caffeine cache using a weak keys for the same goal.


pditommaso:
Add Yourkit thread analysis for the records 

Screenshot 2025-01-07 at 13 59 21

PR #790, opened 2025-01-27
update docs

munishchouhan:
This PR will do the following
1. Current title formatting of metrics.mdx
2. move configuration.md to docs folder
3. change configuration.md to configuration.mdx to use same format for all the docs


docs/configuration.mdx
```diff
@@ -1,10 +1,10 @@
 # Wave Application Configuration
 
-Set Wave configuration values using environment variables or in [`config.yml`](./config.yml) configuration file
+Set Wave configuration values using environment variables or in [`config.yml`](../config.yml) configuration file
```
jason-seqera, line 3: @munishchouhan, I don't think anything outside of the `docs` directory gets included, so this link will be broken.

munishchouhan: changed to absolute url
https://github.com/seqeralabs/wave/blob/master/config.yml

jason-seqera: @munishchouhan, so only files that live inside `docs` will make it into the public docs. If this isn't included, it won't link either. It basically doesn't exist for all intents and purposes.

If it is intended for public docs, it'll need to live inside `docs` as well. If not, we can't link to it either.

munishchouhan: why can't we link this file in our docs?, this is  public url

jason-seqera: There's no prohibition. A complete URL to GitHub definitely works as usual. Sorry.

pditommaso (approved):

PR #821, opened 2025-04-03
Add PostgreSQL based implementation for PersistenceService

pditommaso:
This PR adds an implementation Wave [PeristenceService](https://github.com/seqeralabs/wave/blob/a4cc5199e880cfabacb5a05a16f774740353a6da/src/main/groovy/io/seqera/wave/service/persistence/PersistenceService.groovy#L33-L33) based on PostgreSQL db. 

The overall idea consists on relying on [JSON capability](https://www.postgresql.org/docs/current/datatype-json.html) to storage and retrieve Wave build, scan and mirror information. 

This approach makes it very similar to the current Surreal DB based implementation and it's radically simpler to maintain compared to class relation schema based model.


src/main/groovy/io/seqera/wave/service/persistence/postgres/DbInitService.groovy
```diff
@@ -0,0 +1,88 @@
...
+                id TEXT PRIMARY KEY,
+                data JSONB NOT NULL,
+                created_at TIMESTAMP DEFAULT NOW()
+            );
+
+            CREATE TABLE IF NOT EXISTS wave_scan (
+                id TEXT PRIMARY KEY,
+                data JSONB NOT NULL,
+                created_at TIMESTAMP DEFAULT NOW()
+            );
```
gavinelder, line 56: Question is this timestamp supplied or generated using the default value, if it's the later there may be an optimization we can do in Postgres using  generated columns 

This removes the need for additional JSON parsing at write where we can have an additional benefit of indexing

For example.

```
CREATE TABLE IF NOT EXISTS wave_scan (
    id TEXT PRIMARY KEY,
    data JSONB NOT NULL,
    created_at TIMESTAMP DEFAULT NOW(),
    creation_time TIMESTAMP GENERATED ALWAYS AS ((data->>'creationTime')::timestamp) STORED
);
```

The reference for this is https://www.postgresql.org/docs/current/ddl-generated-columns.html

src/main/groovy/io/seqera/wave/service/persistence/postgres/data/BuildRow.groovy
```diff
@@ -0,0 +1,49 @@
...
+@Introspected
+@CompileStatic
+@Table(name = "wave_build")
+class BuildRow {
+
+    @Id
+    String id
+
+    @TypeDef(type = DataType.JSON)
+    String data  // JSONB data will be retrieved as a string
```
gavinelder, line 45: I am not very familiar with this type of data structure when it comes to Java , reading some items on micronaut is there a particular reason for usig a String type here ? 

Asking for my own education as I was following this guide https://micronaut-projects.github.io/micronaut-data/4.6.0/guide/index.html#sqlJsonType

gavinelder (commented):
Appreciate this is not ready for review , was super interested and this PR really surfaced the underlying data structures wave uses for me. 

Left a comment and question.

src/main/groovy/io/seqera/wave/service/persistence/postgres/data/BuildRow.groovy
pditommaso: I was not considering that because the target data structure should be a [WaveBuildRecord](https://github.com/seqeralabs/wave/blob/ea903fd247c0b47d3080adf989186adbdba38f1f/src/main/groovy/io/seqera/wave/service/persistence/WaveBuildRecord.groovy#L42-L42) object, therefore the idea is to load as a string and then deserialise into a `WaveBuildRecord` here https://github.com/seqeralabs/wave/blob/ea903fd247c0b47d3080adf989186adbdba38f1f/src/main/groovy/io/seqera/wave/service/persistence/postgres/PostgresPersistentService.groovy#L87

However i've just realised that declaring as 

```
    @TypeDef(type = DataType.JSON)
    WaveBuildRecord data
```


It manages the ser-deserialization automatically, neat!

pditommaso:
@munishchouhan can you test this at your convenience?

munishchouhan:
Wave is throwing error:
`Message: Failed to load driver class org.h2.Driver in either of HikariConfig class loader or Thread context classloader`
we have datasource config in dev env and micronaut data is trying to setup h2 and cannot find depednecy

https://github.com/seqeralabs/wave/blob/2648c904a2bad8527db4ad1f6727b18f31548033/src/main/resources/application-dev.yml#L39-L46

Solution:

1. Either we remove config from dev env
2. or we change h2 dependency to runtime: `runtimeOnly 'com.h2database:h2'`

pditommaso:
Don't remember why i've added it in `-dev`, let's try to remove it

munishchouhan:
localDB setup:

```
docker run -d \
  --name wave_postgres \
  -v my_dbdata:/var/lib/postgresql/data \
  -p 5432:5432 \
  -e POSTGRES_USER=postgres \
  -e POSTGRES_PASSWORD=postgres \
  -e POSTGRES_DB=wave \
  postgres:latest
```

pditommaso:
Actually to properly tests this we need a RDS Postgres instance in dev cluster. @seqeralabs/devops can you please assist with this? 

@munishchouhan not sure make sense to commit local db config here https://github.com/seqeralabs/wave/pull/821/commits/11dc6ed429cd4ca0525ffde33af9934da1559cc6

munishchouhan:
> Actually to properly tests this we need a RDS Postgres instance in dev cluster. @seqeralabs/devops can you please assist with this?
> 
> @munishchouhan not sure make sense to commit local db config here [11dc6ed](https://github.com/seqeralabs/wave/commit/11dc6ed429cd4ca0525ffde33af9934da1559cc6)

It is same as we have for surrealdb


> Actually to properly tests this we need a RDS Postgres instance in dev cluster. @seqeralabs/devops can you please assist with this?
> 
> @munishchouhan not sure make sense to commit local db config here [11dc6ed](https://github.com/seqeralabs/wave/commit/11dc6ed429cd4ca0525ffde33af9934da1559cc6)

I made it same as surrealdb config easy for development

pditommaso:
> It is same as we have for surrealdb

The plan is to run as managed db, not embedded as Surreal

bebosudo:
Please reach out in `#team-devops` for details.

munishchouhan:
Tested in Dev with postgres RDS cluster
10 builds and scans
10 mirrors
10 transfers 

No relevant errors

munishchouhan (approved):

PR #822, opened 2025-04-08
Add support for compression mode

pditommaso:
This PR adds support for compression to container build requests.


pditommaso:
It should be considered to add the `compression` to the containerId function to force a re-build when changing the compression mode 

https://github.com/seqeralabs/wave/blob/b078ae0dafa613a95f9eadbffc750ce173979f2f/src/main/groovy/io/seqera/wave/util/ContainerHelper.groovy#L306-L312

PR #830, opened 2025-05-06
Request with a different container config does not have a different container ID

munishchouhan:
This PR shows that a request with a different container config does not have a different container ID.


munishchouhan:
It works in case of freeze because Wave adds config in the containerfile before generating the container ID
https://github.com/seqeralabs/wave/blob/647edc52861d86eb0d64ff640e12846b7b9e2dbf/src/main/groovy/io/seqera/wave/service/builder/FreezeServiceImpl.groovy#L155

PR #840, opened 2025-05-19
added jdbc binder for promethues matrics

munishchouhan:
This pr will expose jdbc metrics to prometheus

### DIFF
--- a/src/main/resources/application-prometheus.yml
+++ b/src/main/resources/application-prometheus.yml
@@ -19,6 +19,8 @@ micronaut:
           enabled: true
         channels:
           enabled: true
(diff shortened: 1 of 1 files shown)


src/main/resources/application-prometheus.yml
```diff
@@ -19,6 +19,8 @@ micronaut:
           enabled: true
         channels:
           enabled: true
+      jdbc :
+        enabled : true
```
pditommaso, line 23:
```suggestion
      jdbc:
        enabled: true
```

nit

pditommaso (approved):

PR #859, opened 2025-06-20
Store wave pull records in DB

munishchouhan:
This PR will create wave_pull table and add entry with requestId and time every time there is a pull.

cc @lialinn 

Note: please ignore the commit message


munishchouhan:
test:

Screenshot 2025-06-20 at 17 08 04

munishchouhan:
tested in local
Screenshot 2025-06-20 at 17 54 28

pditommaso:
Not sure this should be added. These metrics are tracked already via Redis

munishchouhan:
> Not sure this should be added. These metrics are tracked already via Redis 

Currently we only store metrics per org, arch and date. This implementation will provide us pull per every parameter in the request object like image, user, ip etc.

munishchouhan:
> Not sure this should be added. These metrics are tracked already via Redis

This is not just for matrices; it's a record of actions performed through Wave, like build, scan, etc. The client also pulls through Wave.
So, having wave_pull makes sense for that purpose too.

pditommaso:
Closing in favour of registry analytics

munishchouhan:
@pditommaso, as discussed with the analytics team, registry analytics still does not provide the information about wave usage.
So this is still a valid solution to provide wave metrics to product analytics.
cc @lialinn @MichaelTansiniSeqera @gavinelder

PR #866, opened 2025-07-09
Bump k8s java client to 24.0.0

munishchouhan:
This version is required with k8s version 1.31
reason: there ia breaking API change-> user field has been added to V1Containerstatus


pditommaso:
Ok, let's move to ready when you are happy with this PR

VERSION
```diff
@@ -1 +1 @@
-1.22.0
+1.22.2-A0
```
pditommaso:
```suggestion
1.22.0
```

PR #885, opened 2025-08-27
docs: fix wave build steps

gavinelder:
The following updates the relevant build documentation guide & steps. 

The k8s config was not nested properly under the build config. 

Key items. 

- Changed `workspace` from `/build/work` -> `/build/workspace` this is more to help align terminology and a not strictly needed but helpful.
- `build.k8s.storage.mountpath` set a mount on `/build` for the pod so the directory is usable for build pods to prevent directory not found errors. 
- Added various config examples which a consumer may want to use such as resource limits / node selectors etc.

### DIFF
--- a/docs/install/configure-wave-build.md
+++ b/docs/install/configure-wave-build.md
@@ -134,18 +134,30 @@ data:
       # Enable build service
       build:
         enabled: true
-        workspace: '/build/work'
+        workspace: '/build/workspace'
+        # Optional: Retain failed builds to gather logs & inspect
+        cleanup: "OnSuccess"
         # Optional: Configure build timeouts
-        timeout: '1h'
-        # Optional: Configure resource limits for build pods
-        resources:
-          requests:
-            memory: '1Gi'
-            cpu: '500m'
-          limits:
-            memory: '4Gi'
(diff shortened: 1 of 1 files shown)


christopher-hakkaart (approved):
I think this is good to merge. Note that I am working on a PR to refresh a lot of this page and I'll add some text to describe the configuration and optional configuration as a part of that.

docs/install/configure-wave-build.md
```diff
@@ -134,18 +134,30 @@ data:
...
       build:
         enabled: true
-        workspace: '/build/work'
+        workspace: '/build/workspace'
+        # Optional: Retain failed builds to gather logs & inspect
+        cleanup: "OnSuccess"
         # Optional: Configure build timeouts
-        timeout: '1h'
-        # Optional: Configure resource limits for build pods
-        resources:
```
munishchouhan, line 141: why we are removing resources?
this will lead to unpredicted usage by wave build pods

gavinelder: These have been re-added.

munishchouhan: thank you

PR #893, opened 2025-09-02
Wv 217 remove app encoded values for seqera environment

munishchouhan:
This pr will add the following:

1. default profile for config used in Seqera Seqera-hosted wave
2. Add `wave.build.public-repo` in docs


src/main/resources/application-default.yml
pditommaso, line 1: Let's move these into `application-dev.yml`

munishchouhan: done

PR #900, opened 2025-09-04
Added wave.build.storage-type to support different mountpoints

munishchouhan:
this pr will add wave.build.storage-type to support different mountpoints on efs and object storage


pditommaso:
File permissions need to preserved as requirement.

PR #908, opened 2025-09-16
fixed metrics config docs

munishchouhan:
This fill fix the metrics config docs.
we only provide per org and date for now
and also added the link for it.


christopher-hakkaart:
The [Wave Metrics service](https://github.com/seqeralabs/wave/blob/master/docs/metrics.md) page isn't in the Wave sidebar. I think this PR can be merged; however, I'll follow up with a second PR to style the metrics page and a docs repo PR to add it to the sidebar.

PR #921, opened 2025-10-17
docs: Fix error in docs last_update

christopher-hakkaart:
- `last_update` is causing an error with the docs build


llewellyn-sl (approved):
Thanks!

PR #937, opened 2025-11-27
Add configurable singularity init container image

pditommaso:
## Summary
- Add `wave.build.singularity-image-init` configuration setting to customize the init container image used for singularity builds
- Replace hardcoded `busybox` reference with configurable value
- Default to `public.cr.seqera.io/wave/busybox:latest`

## Test plan
- [x] Existing `K8sServiceImplTest` tests pass
- [x] Added validation for init container image in singularity build job spec test

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -53,6 +53,7 @@ wave:
   build:
     buildkit-image: "moby/buildkit:v0.25.2-rootless"
     singularity-image: "public.cr.seqera.io/wave/singularity:v4.2.1-r4"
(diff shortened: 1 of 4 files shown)


src/main/resources/application.yml
```diff
@@ -53,6 +53,7 @@ wave:
   build:
     buildkit-image: "moby/buildkit:v0.25.2-rootless"
```
gavinelder: For the remaining buildkit image it feels like we should align and mirror and serve from `public.cr.seqera.io/wave/buildkit` also.

gavinelder (approved):

PR #945, opened 2025-12-10
Add documentation for bin directory bundling with Wave and Fusion

robsyme:
- Document wave.bundleProjectResources configuration option
- Add "Workflow bin scripts with Fusion" section explaining bundling behavior
- Update fingerprint components list to include container layers
- Add warning about frozen images when Wave is disabled

Resolves examples of confusion when users didn't understand when bin scripts are baked into containers vs staged at runtime.


pditommaso (changes requested):
Thanks Rob, but wave.bundleProjectResources is meant to be an internal only setting to enable intra components communication

PR #1024, opened 2026-04-09
Use micromamba/v2 as default build template for conda packages

pditommaso:
## Summary
- Switch the default build template for CONDA packages from `conda/micromamba:v1` to `conda/micromamba:v2`
- `BuildTemplate.defaultTemplate()` now returns `CONDA_MICROMAMBA_V2` for CONDA package type
- `ContainerHelper.containerFileFromRequest()` routes to v2 by default, v1 only when explicitly requested
- Existing v1 tests updated to explicitly specify `CONDA_MICROMAMBA_V1` template

## Test plan
- [x] Existing unit tests pass (`ContainerHelperTest`, `CondaHelperTest`)
- [ ] Verify conda builds on staging use v2 template by default
- [ ] Verify explicit `conda/micromamba:v1` template still works

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1027, opened 2026-04-15
Fix conda builds for packages requiring __cuda virtual package

### LINKED ISSUE #1026: conda-file builds cannot install packages requiring __cuda virtual package
conda-file builds cannot install packages requiring `__cuda` virtual package

### Summary

Wave's conda-file builds cannot install any conda-forge package that depends on the `__cuda` virtual package, because Wave's build servers do not have GPUs and the templates provide no mechanism to set `CONDA_OVERRIDE_CUDA` during the conda solve step. This affects all GPU-enabled packages on conda-forge from mid-2022 onwards, including PyTorch >= 1.12.1, JAX, TensorFlow, and others.

The standard conda/mamba workaround for building on CPU-only systems is setting `CONDA_OVERRIDE_CUDA=<version>` as an environment variable during the solve. Wave does not support this.

This was previously reported in the community forum ([Problems with GPU-enabled wave images](https://community.seqera.io/t/problems-with-gpu-enabled-wave-images/1596), Jan 2025) by Nico_Trummer, who hit the same `__cuda` resolution failure building JAX GPU containers. That thread is unresolved - Paolo responded about the build timeout aspect ([#597](https://github.com/seqeralabs/wave/issues/597)) but the core `CONDA_OVERRIDE_CUDA` gap was not addressed. This issue documents the problem fully with reproduction steps, tested workarounds, and concrete fix proposals.

### Reproduction

Given this `environment.gpu.yml`:
```yaml
channels:
  - conda-forge
  - bioconda
dependencies:
  - "bioconda::ribodetector=0.3.3"
  - "conda-forge::pytorch-gpu>=2.0"
```

```bash
wave --conda-file environment.gpu.yml --platform linux/amd64 --freeze --await
```

Fails with:
```
Could not solve for environment specs
pytorch-gpu >=2.0 is not installable because it requires
  pytorch [...], which requires
    __cuda =* *, which is missing on the system.
```

Build log: https://wave.seqera.io/view/builds/bd-bbf66b1b68ac0df5_1

### Why this matters

The `__cuda` virtual package requirement was introduced in conda-forge's PyTorch builds starting at version 1.12.1. We verified this by testing every major version on a Linux x86_64 system without `CONDA_OVERRIDE_CUDA` set:

| pytorch-gpu version | Resolves without `__cuda`? | CUDA packaging |
|---|---|---|
| 1.11.0 | Yes | Uses `cudatoolkit` (regular package) |
| 1.12.1 | **No** | Requires `__cuda` virtual package |
| 1.13.1 | **No** | Requires `__cuda` virtual package |
| 2.0.0 | **No** | Requires `__cuda` virtual package |
| 2.5.1 | **No** | Requires `__cuda` virtual package |

This means Wave can only install `pytorch-gpu=1.11.0` (March 2022, CUDA 11.1) via `--conda-file`. All newer versions fail.

This is a real problem in production: the nf-core/rnaseq pipeline's GPU ribodetector test [stalls for 2 hours under Singularity](https://github.com/nf-core/rnaseq/pull/1788) because the old PyTorch 1.11.0 container (CUDA 11.1 runtime) deadlocks on CUDA 12.x hosts. We need a container with modern PyTorch (CUDA 12.x), but Wave's `--conda-file` path cannot build one.

### What we tried

**`--config-env 'CONDA_OVERRIDE_CUDA=12.6'`**: This sets the env var in the final container image, not during the build/solve step. The conda solver never sees it. ([Build log](https://wave.seqera.io/view/builds/bd-566e7b0788a824ff_1))

**`--conda-run-command`**: Per the Wave source (`TemplateUtils.java`, `addCommands` method), this appends Dockerfile `RUN` commands *after* the `micromamba install` step. It cannot influence the conda solve.

**`--conda-base-image nvidia/cuda:12.6.3-base-ubuntu24.04`**: Fails because the NVIDIA base image does not have micromamba installed. ([Build log](https://wave.seqera.io/view/builds/bd-b587fdb7e4ed8bfc_1))

**Adding `__cuda>=12` as a dependency in the YAML**: The solver correctly identifies it as a virtual package that must be provided by the system and refuses to install it. ([Build log](https://wave.seqera.io/view/builds/bd-5a19be48d123eec7_1))

### Workaround: custom Dockerfile (works, but no community freeze)

A custom Dockerfile with `CONDA_OVERRIDE_CUDA` set inline works:

```dockerfile
FROM mambaorg/micromamba:1.5.10-noble
COPY conda.yml /tmp/conda.yml
RUN CONDA_OVERRIDE_CUDA="12.6" micromamba install -y -n base -f /tmp/conda.yml && micromamba clean -a -y
USER root
ENV PATH="$MAMBA_ROOT_PREFIX/bin:$PATH"
```

```bash
wave -f Dockerfile --context . --platform linux/amd64 --await --tower-token $TOKEN
# Returns: wave.seqera.io/wt/51f41f22dbb8/wave/build:96b4265a2148d918
```

We pulled this container and verified it contains PyTorch 2.10.0 with CUDA 12.9 and ribodetector 0.3.3 works correctly:

```
$ docker run --rm <image> python -c "import torch; print(torch.__version__, torch.version.cuda)"
2.10.0 12.9

$ docker run --rm <image> ribodetector --version
ribodetector 0.3.3
```

However, the custom Dockerfile path requires `--build-repo` for `--freeze`, meaning it cannot be frozen to the community Wave registry (`community.wave.seqera.io/library/...`). This makes it unusable for nf-core modules, which rely on community-frozen container URLs.

### Proposed solutions

There are several ways to address this, ranging from minimal to comprehensive:

#### Option A: Add `{{conda_env_prefix}}` template placeholder (minimal, targeted)

Add a new placeholder to the Dockerfile templates that injects environment variables before `micromamba install`:

```diff
 FROM {{mamba_image}} AS build
 COPY --chown=$MAMBA_USER:$MAMBA_USER conda.yml /tmp/conda.yml
-RUN micromamba install -y -n base -f /tmp/conda.yml \
+RUN {{conda_env_prefix}}micromamba install -y -n base -f /tmp/conda.yml \
```

Expose this via a new CLI flag (e.g., `--conda-solve-env KEY=VALUE`) and Nextflow config option. The placeholder would render as `CONDA_OVERRIDE_CUDA="12.6" ` when set, or empty string when not.

This requires changes to:
- All 8 conda templates (v1/v2, Docker/Singularity, file/packages)
- `CondaOpts.java` (new field)
- `TemplateUtils.java` (new binding)
- Wave CLI (new flag)
- Nextflow Wave config (new option)

#### ~~Option B: Set `CONDA_OVERRIDE_CUDA` unconditionally in templates~~

*Ruled out: testing shows that for packages with both CPU and GPU variants (e.g. bare `pytorch` without the `-gpu`/`-cpu` suffix), the override causes the solver to prefer CUDA builds, silently pulling in hundreds of MB of CUDA toolkit into containers that never intended to use a GPU.*

#### Option C: Detect GPU packages and set override automatically

Wave could inspect the conda environment spec for known GPU metapackages (`pytorch-gpu`, `jaxlib`, `tensorflow-gpu`, etc.) or `__cuda` dependencies and automatically set `CONDA_OVERRIDE_CUDA` during the solve. This would be the most user-friendly option but requires more implementation effort.

#### Option D: Two-pass solve with automatic retry

Run the solve normally. If it fails and the error output contains `__cuda`, retry with `CONDA_OVERRIDE_CUDA` set. The template's `RUN` command would become something like:

```bash
micromamba install -y -n base -f /tmp/conda.yml || \
  (micromamba install --dry-run -y -n base -f /tmp/conda.yml 2>&1 | grep -q __cuda \
   && CONDA_OVERRIDE_CUDA="12" micromamba install -y -n base -f /tmp/conda.yml)
```

This requires no new CLI flags, no package list, and no repodata inspection. The solver itself discovers whether `__cuda` is needed. The cost is one extra failed solve (~4s) for GPU environments only; non-GPU environments succeed on the first pass with zero overhead. This could be implemented entirely within the existing templates.

### Related

- [Community thread: Problems with GPU-enabled wave images](https://community.seqera.io/t/problems-with-gpu-enabled-wave-images/1596) (Jan 2025, unresolved)
- [conda-forge docs: CONDA_OVERRIDE_CUDA](https://conda-forge.org/docs/user/tipsandtricks/#installing-cuda-enabled-packages-like-tensorflow-and-pytorch)

pditommaso:
## Summary

- Fixes conda-file and conda-packages builds that fail when a dependency requires the `__cuda` virtual package (e.g., `pytorch-gpu >= 1.12.1`, JAX, TensorFlow)
- Adds retry-on-failure logic to all 4 micromamba v2 templates (Docker/Singularity x file/packages): the first install attempt output is captured, and if it fails with `__cuda` in the error, retries with `CONDA_OVERRIDE_CUDA="12"` set
- Zero overhead for non-GPU builds; GPU builds incur only a fast solver failure (~4s) before the successful retry

Closes #1026

## Test plan

- [x] All 34 TemplateUtilsTest tests pass
- [ ] Verify with a real conda-file build containing pytorch-gpu>=2.0
- [ ] Verify non-GPU builds (e.g., bwa salmon) are unaffected

### DIFF
--- a/src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-file.txt
+++ b/src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-file.txt
@@ -1,6 +1,8 @@
 FROM {{mamba_image}} AS build
 COPY --chown=$MAMBA_USER:$MAMBA_USER conda.yml /tmp/conda.yml
-RUN micromamba install -y -n base -f /tmp/conda.yml \
+RUN (micromamba install -y -n base -f /tmp/conda.yml > /tmp/mamba.log 2>&1 \
+    || (cat /tmp/mamba.log >&2 && grep -q __cuda /tmp/mamba.log \
+        && CONDA_OVERRIDE_CUDA="12" micromamba install -y -n base -f /tmp/conda.yml)) \
     {{base_packages}}
     && micromamba env export --name base --explicit > environment.lock \
     && echo ">> CONDA_LOCK_START" \
--- a/src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-file.txt
+++ b/src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-file.txt
@@ -1,6 +1,7 @@
 FROM {{mamba_image}} AS build
 COPY --chown=$MAMBA_USER:$MAMBA_USER conda.yml /tmp/conda.yml
 RUN (micromamba install -y -n base -f /tmp/conda.yml > /tmp/mamba.log 2>&1 \
+    && cat /tmp/mamba.log \
     || (cat /tmp/mamba.log >&2 && grep -q __cuda /tmp/mamba.log \
         && CONDA_OVERRIDE_CUDA="12" micromamba install -y -n base -f /tmp/conda.yml)) \
     {{base_packages}}
--- a/src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-file.txt
+++ b/src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-file.txt
@@ -3,7 +3,7 @@ COPY --chown=$MAMBA_USER:$MAMBA_USER conda.yml /tmp/conda.yml
 RUN (micromamba install -y -n base -f /tmp/conda.yml > /tmp/mamba.log 2>&1 \
     && cat /tmp/mamba.log \
     || (cat /tmp/mamba.log >&2 && grep -q __cuda /tmp/mamba.log \
-        && CONDA_OVERRIDE_CUDA="12" micromamba install -y -n base -f /tmp/conda.yml)) \
+        && CONDA_OVERRIDE_CUDA="99" micromamba install -y -n base -f /tmp/conda.yml)) \
     {{base_packages}}
     && micromamba env export --name base --explicit > environment.lock \
     && echo ">> CONDA_LOCK_START" \
--- a/src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-packages.txt
+++ b/src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-packages.txt
@@ -3,7 +3,7 @@ RUN \
     (micromamba install -y -n base {{channel_opts}} {{target}} > /tmp/mamba.log 2>&1 \
     && cat /tmp/mamba.log \
     || (cat /tmp/mamba.log >&2 && grep -q __cuda /tmp/mamba.log \
-        && CONDA_OVERRIDE_CUDA="12" micromamba install -y -n base {{channel_opts}} {{target}})) \
+        && CONDA_OVERRIDE_CUDA="99" micromamba install -y -n base {{channel_opts}} {{target}})) \
     {{base_packages}}
     && micromamba env export --name base --explicit > environment.lock \
     && echo ">> CONDA_LOCK_START" \
--- a/src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-packages.txt
+++ b/src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-packages.txt
@@ -1,6 +1,8 @@
 FROM {{mamba_image}} AS build
 RUN \
-    micromamba install -y -n base {{channel_opts}} {{target}} \
+    (micromamba install -y -n base {{channel_opts}} {{target}} > /tmp/mamba.log 2>&1 \
+    || (cat /tmp/mamba.log >&2 && grep -q __cuda /tmp/mamba.log \
+        && CONDA_OVERRIDE_CUDA="12" micromamba install -y -n base {{channel_opts}} {{target}})) \
     {{base_packages}}
     && micromamba env export --name base --explicit > environment.lock \
     && echo ">> CONDA_LOCK_START" \
--- a/src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-packages.txt
+++ b/src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-packages.txt
@@ -1,6 +1,7 @@
 FROM {{mamba_image}} AS build
 RUN \
     (micromamba install -y -n base {{channel_opts}} {{target}} > /tmp/mamba.log 2>&1 \
+    && cat /tmp/mamba.log \
     || (cat /tmp/mamba.log >&2 && grep -q __cuda /tmp/mamba.log \
         && CONDA_OVERRIDE_CUDA="12" micromamba install -y -n base {{channel_opts}} {{target}})) \
     {{base_packages}}
--- a/src/main/resources/templates/conda-micromamba-v2/singularityfile-conda-file.txt
+++ b/src/main/resources/templates/conda-micromamba-v2/singularityfile-conda-file.txt
@@ -4,6 +4,7 @@ From: {{mamba_image}}
     {{wave_context_dir}}/conda.yml /scratch/conda.yml
 %post
     micromamba install -y -n base -f /scratch/conda.yml > /tmp/mamba.log 2>&1 \
+        && cat /tmp/mamba.log \
         || (cat /tmp/mamba.log >&2 && grep -q __cuda /tmp/mamba.log \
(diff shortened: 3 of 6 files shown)


pinin4fjords:
Minor suggestion: consider using a higher sentinel value than `"12"` for `CONDA_OVERRIDE_CUDA`. The value just needs to satisfy `__cuda >=N` constraints - the solver picks the best available CUDA build from the channels regardless. `"12"` works today but would fail if a future package requires `__cuda >=13`. A value like `"99"` is equally correct and future-proof:

```
$ CONDA_OVERRIDE_CUDA=99 micromamba create --dry-run -f env_gpu.yml
  + cuda-version  13.2    # solver still picks the best available
  + pytorch-gpu   2.10.0  cuda129_mkl_...
```

The value is passed through as a raw string with no validation ([`virtual_packages.cpp`](https://github.com/mamba-org/mamba/blob/main/libmamba/src/core/virtual_packages.cpp#L21-L28)).

munishchouhan:
Shouldn't we add it for micromamba-v1 also?

pditommaso:
Uff, you are never happy folks :D

pinin4fjords:
> Does this hard code CUDA 12? CUDA has an annoying dependency matrix so hard coding feels risky (I need 11! I need 13!).

@adamrtalbot The `"12"` doesn't hardcode CUDA 12 in the container - it just tells the solver "assume CUDA >= 12 is available." The solver then picks the best CUDA build from the channels independently (we got `cuda-version 12.9` with pytorch-gpu 2.10.0). If someone's environment needs CUDA 11.8, the solver will select it as long as the packages support it.

That said, using `"99"` instead of `"12"` avoids any ambiguity and future-proofs against `__cuda >=13` constraints.

pditommaso:
Could a fair starting point? we could add a cuda specific option later to specific it

munishchouhan:
> Shouldn't we add it for micromamba-v1 also?

the issue is reported on the default template which is micromamba-v1 
https://wave.seqera.io/view/builds/bd-bbf66b1b68ac0df5_1

So we also need to make changes to the default template

pditommaso:
We are going to switch to v2 soon as default https://github.com/seqeralabs/wave/pull/1024

pinin4fjords:
> Could a fair starting point? we could add a cuda specific option later to specific it

Agreed, this is a great starting point. The two-pass logic is the important part - it unblocks all current GPU builds with zero impact on non-GPU builds. I don't think a user-facing `--cuda` flag would add much value in practice - the override only gates which builds the solver considers eligible, it doesn't pin anything in the container. If someone needs a specific CUDA version, they'd pin `cuda-version=X` in their environment YAML directly.

One thing worth noting: the sentinel value does affect the solve. With `CONDA_OVERRIDE_CUDA=11.8`, the solver excludes CUDA 12+ builds entirely and picks `pytorch-gpu 2.5.1` instead of `2.10.0`. Using a high value like `99` means "let the solver pick the best available," which is the right default for container builds where you don't know the runtime host's driver version.

pinin4fjords:
> We are going to switch to v2 soon as default #1024

I'd love to be able to get good builds ASAP though :-)

munishchouhan:
> We are going to switch to v2 soon as default #1024

But we are still not removing micromamba-v1, so worth to add the fix there too

pditommaso:
@pinin4fjords so you are ok to merge or you want `CONDA_OVERRIDE_CUDA=11.8` ?

pinin4fjords:
> @pinin4fjords so you are ok to merge or you want `CONDA_OVERRIDE_CUDA=11.8` ?

I'd suggest `CONDA_OVERRIDE_CUDA="99"` rather than `11.8` or `12`. It's clearly not a real CUDA version, so nobody will mistake it for a pin, and it's high enough that it'll satisfy any future `__cuda >=N` constraint without needing to be updated. The solver picks the best available build from the channels regardless of this value.

Also +1 to @munishchouhan's point about adding the fix to the v1 templates too - looks like they're not included in the PR yet.

pditommaso:
Done, let's test in v2 before applying in v1 template

pditommaso:
Can you please review and approve ?

pinin4fjords (approved):
Looks good. The two-pass logic is correct, `99` is the right sentinel value, and testing in v2 first before applying to v1 makes sense. Thanks for the quick turnaround on this.

PR #1029, opened 2026-04-17
feat: Make server listening port configurable via WAVE_SERVER_PORT mn env var

bebosudo:
Allows users to configure the wave server listening port using the explicit variable `WAVE_SERVER_PORT` instead of using the implicit micronaut variable `MICRONAUT_SERVER_PORT`.


gavinelder (approved):

src/main/resources/application.yml
```diff
@@ -57,7 +57,7 @@ micronaut:
 wave:
   allowAnonymous: true
   server:
-      url: "${WAVE_SERVER_URL:`http://localhost:9090`}"
+      url: "${WAVE_SERVER_URL:`http://localhost:${WAVE_SERVER_PORT:9090}`}"
```
pditommaso: I'm not really sure WAVE_SERVER_PORT within backtick is replaced by the variable

bebosudo: Indeed. That's the reason tests are failing

pditommaso (approved):

munishchouhan (approved):

bebosudo:
Tests passing, I'll merge it

PR #1039, opened 2026-05-05
chore: port wave CI to push to internal ECR SEC-1395

ShahzebMahmood:
## Summary
- ported wave CI to push to the new internal ECR alongside the legacy one
- tag-and-push.sh switched from jib (push) to jibDockerBuild (local) so the workflow handles all pushes
- legacy pushes (wave/app and nf-tower-enterprise/wave) preserved on [release] commits; new internal pushes go to internal/seqeralabs/wave/app and platform/wave

PR #1046, opened 2026-05-19
fix: port wave CI release flow with non-enterprise jibRepo fix SEC-1395

ShahzebMahmood:
re-applies #1039 (reverted in 7ad68e94) with the bug fix that caused the revert.

**what broke**: the non-enterprise gradle call was missing -PjibRepo, so JIB tagged the image as `wave/app:$version` (no v prefix) but the script then tried `docker tag wave/app:$TAG server:$TAG` where $TAG is `v$version`. mismatch, release run 25995019680 died.

**fix**: per @bebosudo's suggestions, build the image once with a generic name (`wave/server:$TAG`) and let the workflow handle ECR retagging/pushing. dropped the internal ECR push since wave doesn't need internal images.

gating uses paolo's detect-release output from ff8c7af80.

SEC-1395


ShahzebMahmood:
@claude

tag-and-push.sh
```diff
@@ -42,10 +42,15 @@ if [[ $RELEASE ]]; then
...
-  ./gradlew -PjibRepo=195996028523.dkr.ecr.eu-west-1.amazonaws.com/nf-tower-enterprise/wave:$TAG jib
+  # Build images locally (jibDockerBuild loads to the local docker daemon).
+  # Push happens from the workflow so we can also push to the new internal/enterprise ECR repos.
+  echo "Building non-enterprise image $TAG"
+  ./gradlew -PjibRepo=195996028523.dkr.ecr.eu-west-1.amazonaws.com/wave/app:$TAG jibDockerBuild
+  docker tag 195996028523.dkr.ecr.eu-west-1.amazonaws.com/wave/app:$TAG server:$TAG
+
+  echo "Building enterprise image $TAG"
+  ./gradlew -PjibRepo=195996028523.dkr.ecr.eu-west-1.amazonaws.com/nf-tower-enterprise/wave:$TAG jibDockerBuild
+  docker tag 195996028523.dkr.ecr.eu-west-1.amazonaws.com/nf-tower-enterprise/wave:$TAG wave/server:$TAG
```
bebosudo: We can skip the push to the internal repo since the wave application doesn't seem to use it, probably because it's public.

We could just name the image `wave/server:$TAG` and perform the retagging later for the legacy ECR in a separate step inside the GHA, so when we'll clean up the legacy ECR we won't need to edit the tag-and-push script again

```suggestion
  ./gradlew -PjibRepo=wave/server:$TAG jibDockerBuild
```

.github/workflows/build.yml
```diff
@@ -116,6 +119,131 @@ jobs:
...
+          username: ${{ secrets.TOWER_CI_AWS_ACCESS }}
+          password: ${{ secrets.TOWER_CI_AWS_SECRET }}
+        env:
+          AWS_REGION: eu-west-1
+
+      - name: Push images to legacy ECR
+        if: "steps.detect-release.outputs.is_release == 'true'"
+        run: |
+          TAG="${{ steps.release.outputs.version }}"
+          docker push 195996028523.dkr.ecr.eu-west-1.amazonaws.com/wave/app:$TAG
```
bebosudo:
```suggestion
          docker tag wave/server:$TAG 195996028523.dkr.ecr.eu-west-1.amazonaws.com/wave/app:$TAG
          docker tag wave/server:$TAG 195996028523.dkr.ecr.eu-west-1.amazonaws.com/nf-tower-enterprise/wave:$TAG
          docker push 195996028523.dkr.ecr.eu-west-1.amazonaws.com/wave/app:$TAG
```

Do we even need to keep pushing to the `/wave/app` repo on the legacy ECR since it's the same image we push to the enterprise repo?

.github/workflows/build.yml
```diff
@@ -116,6 +119,131 @@ jobs:
...
+
+      - name: Push image to internal registry
+        if: "steps.detect-release.outputs.is_release == 'true'"
+        shell: bash
+        run: |
+          LOCAL=server:${{ steps.release.outputs.version }}
+          REMOTE="${{ steps.login-ecr-internal.outputs.registry }}/internal/${{ github.repository }}/${LOCAL}"
+          docker tag "$LOCAL" "$REMOTE"
+          docker push "$REMOTE"
+
```
bebosudo:
```suggestion
```

this can go, wave doesn't need internal images

bebosudo (commented):
Some suggestions

bebosudo (approved):
LGTM
Could you check with the team if we could stop pushing to the `wave/app` legacy repo, since it should contain the same exact images as the enterprise repo?
Internally we're using it for our installations, but we could point our apps to the enterprise repo. We could then remove that in a follow up PR

.github/workflows/build.yml
ShahzebMahmood: @pditommaso @munishchouhan do we still need to push to the /wave/app repo on the legacy ECR, since it's the same image we push to the enterprise repo? happy to drop it if no one's pulling from /wave/app anymore.

pditommaso: I'm lost with these changes. What's supposed to be the legacy ecr ?

ShahzebMahmood: The legacy ECR is `195996028523.dkr.ecr.eu-west-1.amazonaws.com`, where we've historically pushed wave images to `/wave/app` and `/nf-tower-enterprise/wave`. 

We're moving images to the new central ECR `232933512461.dkr.ecr.eu-west-2.amazonaws.com` in eu-west-2 going forward, so trying to figure out if `/wave/app` on the legacy one is still being pulled by anything or if we can drop it

munishchouhan: then first we need to change the repo in wave manifest
https://github.com/seqeralabs/platform-deployment/blob/master/applications/wave/environments/tower-prod/kustomization.yaml

bebosudo: let's get this PR merged so we start pushing images to the new ECR and make the app use them, then we can open a new PR to stop pushing to the `wave/app` repo on the legacy ECR

pditommaso (changes requested):
The rationale of the tag-and-push.sh script is making possible to test the logic without outside Github action. 

In this form makes no sense any more: either 1) remove it completely, 2) implement the full logic into the bash script

ShahzebMahmood:
> The rationale of the tag-and-push.sh script is making possible to test the logic without outside Github action.
> 
> In this form makes no sense any more: either 1) remove it completely, 2) implement the full logic into the bash script

@pditommaso sounds good, in c2279c64. moved the docker tag/push for both ECRs out of build.yml and back into tag-and-push.sh so the full publish flow lives in the script and can be run locally with docker login to both registries.

pditommaso:
Nice, have tried to validate locally, so we are sure it's not breaking next release?

ShahzebMahmood:
ran it locally,  jibDockerBuild with -PjibRepo=wave/server:vTEST builds and loads the image into local docker as `wave/server:vTEST` docker tag for all 3 ECR targets `legacy wave/app, legacy nf-tower-enterprise/wave`, central wave/server) all resolve to the same digest. skipped the actual docker push to keep ECR clean.

the GHA-only parts OIDC + AssumeRole + unset AWS_SESSION_TOKEN before publish.sh can't be exercised outside CI so those will only get verified on the next release run.

image

pditommaso:
Let's make a fake release changing the VERSION to `1.33.6-TEST` and pushing the commit with `[release]` to validate the flow end to end 

https://github.com/seqeralabs/wave/blob/master/VERSION

.github/workflows/build.yml
bebosudo, line 27:
```suggestion
    if: "github.event_name == 'push' || github.repository != github.event.pull_request.head.repo.full_name"
```

[`github.event` is an object](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context) so it'll never evaluate to the string `push`, so the first condition of the check will never pass.
The second condition of the check is also misleading, but I'll open a separate PR for that

.github/workflows/build.yml
```diff
@@ -98,12 +99,40 @@ jobs:
...
+        with:
+          registry: 195996028523.dkr.ecr.eu-west-1.amazonaws.com
+          username: ${{ secrets.TOWER_CI_AWS_ACCESS }}
+          password: ${{ secrets.TOWER_CI_AWS_SECRET }}
+        env:
+          AWS_REGION: eu-west-1
+
+      - name: Configure AWS credentials for platform pusher
+        if: "steps.detect-release.outputs.is_release == 'true'"
+        uses: aws-actions/configure-aws-credentials@ec61189d14ec14c8efccab744f656cffd0e33f37 # ratchet:aws-actions/configure-aws-credentials@v6.1.0
```
bebosudo: Could you give it another pass with ratchet? there have been a few newer releases

bebosudo (commented):
I'd prefer to move away from the `tag-and-push` script, and only keep the build and push logic inside GHA

ShahzebMahmood:
Hey @bebosudo, did what you said. Pulled tag-and-push into the workflow and kept publish.sh for maven. Ran ratchet over the workflow too, all action pins are clean.

Two bugs on the original PR:
1. Role ARN was platform-ecr-pusher which only trusts platform+groundswell, not wave. Swapped to wave-ecr-pusher.
2. wave-ecr-pusher only trusts master so I temp added the branch to its trust to test.

v1.33.6-TEST3 just went green. Same digest pushed to wave/app + nf-tower-enterprise/wave in legacy ECR and wave/server in central.

VERSION needs bumping back before merge.

.github/workflows/build.yml
bebosudo: I see release v6.1.2 is out with a few fixes

VERSION
```diff
@@ -1 +1 @@
-1.33.6
+1.33.6-TEST3
```
bebosudo:
```suggestion
1.33.6
```
let's revert

bebosudo (approved):
LGTM, but I think `ratchet upgrade` should find some newer releases

ShahzebMahmood:
Hey @pditommaso, ready for another look after your feedback.

Worked through this with Alberto. The OIDC role for the new central ECR can only be assumed in CI, not locally, so we moved build + push into the workflow instead of keeping them in tag-and-push.sh. Tested end-to-end with v1.33.6-TEST3, same digest landed in wave/app, nf-tower-enterprise/wave (legacy) and wave/server (central). VERSION back to 1.33.6, ratchet upgrade on the action pins too.

pditommaso (approved):

PR #1059, opened 2026-06-02
feat(ci): use OIDC role for maven publish DEVOPS-1455

ShahzebMahmood:
## Summary
- Swaps the maven publish in `build.yml`'s Release step from static `TOWER_CI_AWS_*` keys to OIDC role assumption.
- Assumes `arn:aws:iam::195996028523:role/WaveMavenPublisherRole` (created in seqeralabs/infrastructure#1549, already applied). Role is scoped tightly to `s3://maven.seqera.io/releases/*` and trusts only `seqeralabs/wave:ref:refs/heads/master`.
- Drops the `unset AWS_SESSION_TOKEN` hack (no longer needed — session token is now intentional and matches the role's region).
- Removes `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` env vars from the Release step.

The `TOWER_CI_AWS_*` GHA secrets stay in place for now — they're still used by the Tests step (`make check`) and the legacy ECR docker login. Those are separate consumers and out of scope.

DEVOPS-1455

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -121,16 +121,20 @@ jobs:
         if: "steps.detect-release.outputs.is_release == 'true'"
         uses: aws-actions/amazon-ecr-login@fa648b43de3d4d023bcb3f89ed6940096949c419 # ratchet:aws-actions/amazon-ecr-login@v2.1.5
 
+      - name: Configure AWS credentials for maven publish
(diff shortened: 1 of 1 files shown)


ShahzebMahmood:
@claude

bebosudo (approved):

PR #1063, opened 2026-06-25
Use conda-lock to generate environment.lock (closes #975)

### LINKED ISSUE #975: Add conda-lock library to generate full conda-lock files for default template
related to #973 and #971, we at nf-core have `environment.yml`s with pip dependencies, e.g. [environment.yml](https://github.com/mashehu/drugresponseeval/blob/a09ea0c1ee03728b61293b2e73ddf7ad3db28ca4/modules/local/collect_results/environment.yml) and currently the micromamba templates don't include non-conda dependencies, see https://github.com/mamba-org/mamba/pull/3623.

Would it be possible to export the lock file with `conda-lock lock -f /tmp/environment.yml` instead of the current `micromamba env export --name base --explicit`?

munishchouhan:
## Summary
- Switch the conda-file build templates (v1 and v2, dockerfile and singularity) from `micromamba env export --name base --explicit` to `conda-lock` for generating `environment.lock`.
- `micromamba env export --explicit` silently omits any `pip:` entries from `environment.yml`, leaving users with mixed conda+pip environments holding an incomplete lock file. `conda-lock` understands `pip` sections and includes them in the lockfile.
- The actual `micromamba install` step that materializes the base env is left untouched, so image contents are produced exactly as before — conda-lock is used only to render the lockfile (so this is a minimal-risk change to image behavior).

## What changed
- `src/main/resources/templates/conda-micromamba-v1/dockerfile-conda-file.txt`
- `src/main/resources/templates/conda-micromamba-v1/singularityfile-conda-file.txt`
- `src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-file.txt`
- `src/main/resources/templates/conda-micromamba-v2/singularityfile-conda-file.txt`
- Test fixtures in `TemplateUtilsTest` and `ContainerHelperTest` updated for the new template output.

The `conda-packages` variants (inline package lists / lock-URL flow) are not changed: they don't accept a conda.yml and inline CLI args can't carry `pip:` deps, so conda-lock offers no benefit there.

## Test plan
- [x] `TemplateUtilsTest` passes (34 tests)
- [x] `ContainerHelperTest` passes (107 tests)
- [ ] Verify a real build with a mixed conda+pip `environment.yml` produces a lockfile containing the pip packages
- [ ] Confirm `CONDA_LOCK_START` / `CONDA_LOCK_END` extraction in `BuildLogServiceImpl` still works with the conda-lock unified lockfile format

Closes #975

### DIFF
--- a/src/main/resources/templates/conda-micromamba-v1/dockerfile-conda-file.txt
+++ b/src/main/resources/templates/conda-micromamba-v1/dockerfile-conda-file.txt
@@ -3,7 +3,7 @@ COPY --chown=$MAMBA_USER:$MAMBA_USER conda.yml /tmp/conda.yml
 RUN micromamba install -y -n base -f /tmp/conda.yml \
     {{base_packages}}
     && micromamba install -y -n base conda-forge::conda-lock \
-    && conda-lock --file /tmp/conda.yml --platform linux-64 --lockfile environment.lock \
+    && micromamba run -n base conda-lock --file /tmp/conda.yml --platform linux-64 --lockfile environment.lock \
     && echo ">> CONDA_LOCK_START" \
     && cat environment.lock \
     && echo "<< CONDA_LOCK_END" \
--- a/src/main/resources/templates/conda-micromamba-v1/dockerfile-conda-file.txt
+++ b/src/main/resources/templates/conda-micromamba-v1/dockerfile-conda-file.txt
@@ -2,7 +2,8 @@ FROM {{base_image}}
 COPY --chown=$MAMBA_USER:$MAMBA_USER conda.yml /tmp/conda.yml
 RUN micromamba install -y -n base -f /tmp/conda.yml \
     {{base_packages}}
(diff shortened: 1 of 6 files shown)


cristianrcv (approved):

pditommaso (changes requested):
This may go into into a new template version

pditommaso:
Related https://github.com/seqeralabs/wave/pull/1064

munishchouhan:
> This may go into into a new template version

ok sure i will create v3

pditommaso:
Let's discuss first what's the best approach for this

munishchouhan:
> Let's discuss first what's the best approach for this

I think we have three options:
1. Changing the current templates - This will affect the current user experience
2. Creating new template version like V3- In this case maintainace will increase because if there is any change in v2 we have to do the same change in v3
3. Add new place holder {{lockfile_command}}- we will make the current conda lock export command as default

PR #1070, opened 2026-07-03
fix: make `which` available for R (bioconda) post-links in conda v2 builds

pditommaso:
## Problem

Conda containers with R/bioconductor packages (e.g. the nf-core rnaseq `TXIMETA_TXIMPORT` step, which installs `bioconductor-tximeta`) build successfully but fail at **runtime**:

```
Loading required package: GenomeInfoDb
Error: package or namespace load failed for 'GenomeInfoDb':
  there is no package called 'GenomeInfoDbData'
```

## Root cause

Bioconductor **data packages** (`GenomeInfoDbData`, etc.) don't ship their payload in the conda package — a conda **post-link** script (`post-link.sh` → `installBiocDataPackage.sh`) downloads it and runs `R CMD INSTALL` at install time.

That `R CMD INSTALL` starts an R session, and bioconda `r-base` shells out to **`/usr/bin/which`** (absolute path) during `utils` `.onLoad` — it ignores `$PATH` and `$R_WHICH`. The v2 build image `mambaorg/micromamba:2-amazon2023` has **no `/usr/bin/which`**, so R can't start, the post-link fails, and the data package is never installed. The post-link failure is **non-fatal** to `micromamba install`, so the build "succeeds" and produces a broken image.

This is why simply installing `conda-forge::which` into `/opt/conda/bin` was **not** enough — R only looks at `/usr/bin/which`.

## Fix

In the conda **micromamba-v2** build stage (as root), symlink the conda `which` to `/usr/bin/which` **before** the user's packages are installed, so bioconda post-link scripts can run R:

```dockerfile
FROM {{mamba_image}} AS build
USER root
...
RUN micromamba install -y -n base conda-forge::which \
    && ln -sf "$MAMBA_ROOT_PREFIX/bin/which" /usr/bin/which \
    && (micromamba install -y -n base -f /tmp/conda.yml ...) ...
```

Applied to all four `conda-micromamba-v2` templates (dockerfile + singularityfile, conda-file + conda-packages). The pixi templates also get `which` as a base package for consistency, though their `*-noble` build image already provides a system `which`. The legacy micromamba **v1** template is unaffected (its Debian-based image ships `which`).

## Verification

Reproduced and fixed locally with real Docker builds on `amazon2023`:
- Before: `R CMD INSTALL` post-link fails (`.onLoad failed ... error in running command`), `GenomeInfoDbData` absent, runtime load fails.
- After: post-link succeeds; final image runs `library(SummarizedExperiment); library(tximport)` → OK.

Unit tests updated: `TemplateUtilsTest`, `ContainerHelperTest` (+ `CondaHelperTest`/`PixiHelperTest` contains-checks) pass.

Refs: https://community.seqera.io/t/which-is-missing-in-recent-seqera-containers/2718

### DIFF
--- a/src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-file.txt
+++ b/src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-file.txt
@@ -1,6 +1,10 @@
 FROM {{mamba_image}} AS build
+USER root
 COPY --chown=$MAMBA_USER:$MAMBA_USER conda.yml /tmp/conda.yml
-RUN (micromamba install -y -n base -f /tmp/conda.yml > /tmp/mamba.log 2>&1 \
+# expose `which` at /usr/bin/which for R (bioconda) post-link scripts; the amazon2023 base image lacks it
+RUN micromamba install -y -n base conda-forge::which \
+    && ln -sf "$MAMBA_ROOT_PREFIX/bin/which" /usr/bin/which \
+    && (micromamba install -y -n base -f /tmp/conda.yml > /tmp/mamba.log 2>&1 \
     && cat /tmp/mamba.log \
     || (cat /tmp/mamba.log >&2 && grep -q __cuda /tmp/mamba.log \
         && CONDA_OVERRIDE_CUDA="99" micromamba install -y -n base -f /tmp/conda.yml)) \
--- a/src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-packages.txt
+++ b/src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-packages.txt
@@ -1,6 +1,10 @@
 FROM {{mamba_image}} AS build
+USER root
+# expose `which` at /usr/bin/which for R (bioconda) post-link scripts; the amazon2023 base image lacks it
 RUN \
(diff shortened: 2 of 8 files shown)


src/main/groovy/io/seqera/wave/util/CondaHelper.groovy
```diff
@@ -121,6 +125,23 @@ class CondaHelper {
...
+     * Ensure the {@code which} command is part of the base packages. The v2 and pixi
+     * templates build on minimal base images (e.g. {@code ubuntu:24.04}) that do not
+     * provide {@code which}, which Nextflow task wrappers require. Installing it as a
+     * conda package guarantees it is present regardless of the base image, whereas the
+     * server-side default packages are bypassed when the client supplies its own.
+     *
+     * @param basePackages The current base packages (may be null/empty)
+     * @return The base packages including {@code conda-forge::which}
+     */
+    static String withWhichPackage(String basePackages) {
```
munishchouhan: Should we check all required packages instead of just `which`?

pditommaso: Apparently is `which` the missing one

munishchouhan: But I think having a method just to check a specific package will lead to redundancy if we add more required packages

munishchouhan (approved):

PR #1072, opened 2026-07-06
fix: gate wave release on master and release-* branches DEVOPS-1531

ShahzebMahmood:
## Summary

A `[release]` commit now only triggers the release steps when pushed from `master` or a `release-*` branch. On any other branch it logs a warning and skips, instead of failing at the `wave-ecr-pusher` OIDC assume.

Run 28649252557 failed with `Not authorized to perform sts:AssumeRoleWithWebIdentity` because a `[release] Bump 1.35.1-B1` commit was pushed from the feature branch `add-which-tool-conda-templates`. The `wave-ecr-pusher` trust only allows `master` and `release-*` (which is correct), so the assume was rightly rejected. The real issue is that `detect-release` fired a full release from any branch with `[release]` in the commit.

This aligns the workflow trigger with the existing OIDC trust. No infra change needed.

DEVOPS-1531


pditommaso:
Is this really needed? we need to test features with temporary releases, and this constraint adds a lot of friction

gavinelder:
The intent is to prevent internal releases and beta releases surfacing in customer deployed locations and limit their use in production environment.

There should still be a retained capability to cut a container from a branch for use in development environments for testing and validation.

@ShahzebMahmood can you look at this PR with that context.

We are trying to ensure the public release of wave artifacts to enterprise customers and consumers meet our quality bar. 

We are not trying to require a merge before container build for testing and development, for example pr-preview image creation. 

having a tag such as `[preview]` which mirrors `[release]` which pushes to dev could be a potential approach.

ShahzebMahmood:
> The intent is to prevent internal releases and beta releases surfacing in customer deployed locations and limit their use in production environment.
> 
> There should still be a retained capability to cut a container from a branch for use in development environments for testing and validation.
> 
> @ShahzebMahmood can you look at this PR with that context.
> 
> We are trying to ensure the public release of wave artifacts to enterprise customers and consumers meet our quality bar.
> 
> We are not trying to require a merge before container build for testing and development, for example pr-preview image creation.
> 
> having a tag such as `[preview]` which mirrors `[release]` which pushes to dev could be a potential approach.

Yes, working on this

ShahzebMahmood:
Reworked this based on your feedback. Two tags now:

`[preview]` builds from any branch and pushes to the internal dev ECR only, tagged with the commit sha so previews don't overwrite each other. No merge needed, so it covers the pr-preview and testing case.

`[release]` is the customer path. Only on master or a release-* branch does it push to the enterprise repo, publish maven, and cut the git tag and release. Anything else just warns and pushes nothing.

Let me know if that matches what you had in mind.

ShahzebMahmood:
@claude

ShahzebMahmood:
@claude

bebosudo (approved):
LGTM!

PR #1073, opened 2026-07-07
Reduce response contents to remove unneeded fields

gavinelder:
## Summary
- Strip user PII (name and email) from the container-detail API responses while **keeping the existing response shapes unchanged** (non-breaking for API clients)
- `GET /v1alpha2/container/{requestId}` still returns `WaveContainerRecord`; the user is reduced to an id-only copy via `WaveContainerRecord.withoutUserDetails()`
- `DescribeWaveContainerResponse.RequestInfo` keeps the `User`-typed `user` field but populates only the id via the new `User.ofId(...)` factory
- Align the build, mirror, and container view pages to show the user id instead of the user name/email
- Mail service is intentionally unaffected — it reads the live request identity and still sends to `user.email`

## Notes
- PII reduction is scoped to the API response builders only; the `User` class itself is unchanged (no `@JsonIgnore`), so name/email remain available to the mail service and internal flows
- Because Jackson is configured with `NON_ABSENT`, an id-only user serializes as `{"id":...}` — same object shape, no name/email leaked
- TypeSpec route/model reverted to `WaveContainerRecord` (no new response model), matching the unchanged wire shape

## Tests
- `./gradlew generateOpenApi :test --tests 'io.seqera.wave.exchange.DescribeWaveContainerResponseTest' --tests 'io.seqera.wave.controller.ContainerControllerTest.should return the container record' --tests 'io.seqera.wave.controller.ViewControllerTest.should create build binding' --tests 'io.seqera.wave.controller.ViewControllerTest.should render build page' --tests 'io.seqera.wave.controller.ViewControllerTest.should render container view page' --tests 'io.seqera.wave.controller.ViewControllerTest.should render in progress build page' --tests 'io.seqera.wave.controller.ViewControllerTest.should render mirror page'`


pditommaso (approved):
Reviewed for breaking changes: verified none of the Wave ecosystem clients (nextflow nf-wave, wave-cli, platform, sched, seqera-mcp) consume the two affected describe endpoints — all use submit + status/build/inspect paths that don't carry the stripped user fields. PII reduction is non-breaking for API clients. Applied minor rename withoutUserDetails() → withUserIdOnly() for clarity. LGTM.

pditommaso:
Small follow-up: I pushed a one-line rename `withoutUserDetails()` → `withUserIdOnly()` — it reads symmetrically with the `User.ofId(...)` it calls and states what actually survives (the id) rather than what's removed. No behavior change. Feel free to revert if you prefer the original. 🙂

gavinelder:
LGTM on the change

Issue #123, opened 2022-09-12
Add ability to specify different rate limit per auth and anonymous users 


pditommaso:
The goal of this ticker is to provide the ability to define two different rate limits both for authenticated users and anonymous users, both for build and pull requests 

Ideally, those settings should be provided by the configuration file as shown below

```
rate-limit:
  build:
    anonymous: "10 / 1 hour"
    authenticated: "10 / 1 min" 
  pull:
    anonymous: "100 / 1 hour"
    authenticated: "100 / 1 min" 
```

It would be fair to include IP address of the requesting user when the request originates from an anonymous user, otherwise, multiple request from different anonymous users would receive the same rate-limit key.

PR #1076, opened 2026-07-07
Support an authenticating egress proxy for outbound HTTP clients

ewels:
## Summary

Addresses https://github.com/nextflow-io/nextflow/issues/7305.

An on-prem Wave deployment behind a corporate egress proxy could not reach container registries or Seqera Platform: the outbound `java.net.http.HttpClient` instances created by `HttpClientFactory` had no proxy selector and — crucially — no proxy authenticator (`java.net.http.HttpClient` ignores `Authenticator.setDefault()`; the authenticator must be set on the client builder).

This PR adds egress proxy support to `HttpClientFactory`, the single chokepoint through which all outbound clients are created (`ProxyClient`, `StreamServiceImpl`, registry auth/lookup, Tower connector, etc.). It mirrors the approach used to fix the same problem in Nextflow (https://github.com/nextflow-io/nextflow/pull/7331): the proxy selector and proxy-scoped authenticator come from the shared **`io.seqera:lib-httpx`** library (`HxProxyConfig`), so both products share the same proxy semantics.

## Changes

- **`io.seqera:lib-httpx:2.6.0`** (new dependency): provides `HxProxyConfig` — per-protocol HTTP/HTTPS proxy entries with a `ProxySelector` honouring the no-proxy list, and an `Authenticator` scoped to `RequestorType.PROXY` and the configured proxy host/port so credentials are only ever released to the proxy itself.
- **`ProxyConfig`** (new, modelled on Nextflow's `nextflow.util.ProxyConfig`): parses `[http[s]://][user:pass@]host[:port]` proxy URIs (percent-encoded credentials supported, with `+` shielded from the form-decoding rule) and resolves the `HxProxyConfig` from either an explicit URI or the `HTTPS_PROXY`/`HTTP_PROXY`/`NO_PROXY` environment variables (per-protocol proxies supported via env vars).
- **`HttpClientConfig`** (now a `@Context` bean): owns the new settings `wave.httpclient.proxy.uri`, `.username`, `.password`, `.no-proxy` alongside the existing `wave.httpclient.*` options, resolves them at startup — falling back to the environment variables when not set — and applies the result to `HttpClientFactory`.
- **`HttpClientFactory`**: clients are built with `HxProxyConfig.toProxySelector()`/`.toAuthenticator()` when a proxy is configured, via a `setProxyConfig(...)` bridge that invalidates the cached clients. With no proxy configured, no `.proxy(...)` call is made and behaviour is unchanged (requests still route via `ProxySelector.getDefault()`, i.e. `-Dhttps.proxyHost` et al. keep working).
- **Docs**: new *Egress proxy* section in `docs/install/reference.md`, including an explicit statement of scope (the Wave process's internal HTTP clients; Micronaut declarative clients, AWS SDK clients, and out-of-process build/scan/blob-cache jobs have their own proxy configuration paths).

No-proxy semantics follow lib-httpx (same as Nextflow): bare host names match the host and its sub-domains, `.domain`/`*.domain` suffixes match sub-domains only, `*` disables proxying, and loopback targets always bypass the proxy. CIDR entries are not supported.

## The `disabledSchemes` dependency

By default the JDK ships `jdk.http.auth.tunneling.disabledSchemes=Basic`, which silently blocks Basic authentication on the HTTPS CONNECT tunnel — i.e. exactly the path used to reach `https://` registries through a proxy. When a proxy **with credentials** is configured, `HttpClientConfig` defaults both `jdk.http.auth.tunneling.disabledSchemes` and `jdk.http.auth.proxying.disabledSchemes` to empty at startup, **unless the operator has already set them**, in which case the operator's value wins.

Because these properties are read once by the JDK HTTP stack, setting them explicitly on the JVM is the belt-and-braces alternative for containerised deploys, documented in `docs/install/reference.md`:

```
JAVA_TOOL_OPTIONS="-Djdk.http.auth.tunneling.disabledSchemes= -Djdk.http.auth.proxying.disabledSchemes="
```

## Tests

- `ProxyConfigTest`: URI/credential parsing, env-var resolution (per-protocol proxies, lowercase variants, credentials), no-proxy matching via the selector (suffix/wildcard/loopback), authenticator scoping (requestor type, host, port).
- `HttpClientFactoryProxyTest`: integration test against an in-process authenticating proxy (`FakeProxyServer`, supporting Basic auth, plain forwarding and CONNECT tunnelling), with target servers bound to a non-loopback address since loopback targets always bypass the proxy:
  - 407 surfaced when the proxy requires auth and no credentials are configured — both redirect policies;
  - successful authenticated requests through the proxy — both redirect policies;
  - no-proxy list bypass;
  - HTTPS CONNECT tunnelling with Basic proxy auth against a local TLS server (self-signed keystore generated on the fly with `keytool`), which requires `-Djdk.http.auth.tunneling.disabledSchemes=` — added to the `test` task in `build.gradle`.

All 44 new tests pass. Pre-existing network-dependent suites fail identically with and without this change in the sandbox environment (registry credentials/egress policy), i.e. no regression.

## Backwards compatibility

No proxy configured (neither settings nor environment variables) → no behaviour change. Note that deployments that already export `HTTPS_PROXY`/`HTTP_PROXY` will now have those honoured by Wave's outbound clients; `NO_PROXY` is honoured accordingly.

https://claude.ai/code/session_018zJyj9FkzUpMNuH51ScqAQ

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -167,7 +167,7 @@ jobs:
       - name: Login to internal Amazon ECR
         id: login-ecr-internal
         if: "steps.detect-release.outputs.is_preview == 'true'"
-        uses: aws-actions/amazon-ecr-login@d539f0932e70871a027e9d5a9d8fc38589180a64 # v2.1.6
+        uses: aws-actions/amazon-ecr-login@03f1aad4c6c7ffd436567f42f9384779290529bd # v2.1.7
 
       - name: Push image to internal ECR
         if: "steps.detect-release.outputs.is_preview == 'true'"
@@ -203,7 +203,7 @@ jobs:
 
       - name: Login to enterprise Amazon ECR
         if: "steps.detect-release.outputs.is_publish == 'true'"
-        uses: aws-actions/amazon-ecr-login@d539f0932e70871a027e9d5a9d8fc38589180a64 # v2.1.6
+        uses: aws-actions/amazon-ecr-login@03f1aad4c6c7ffd436567f42f9384779290529bd # v2.1.7
 
       - name: Configure AWS credentials for maven publish
         if: "steps.detect-release.outputs.is_publish == 'true'"
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -41,7 +41,7 @@ jobs:
           fetch-depth: 1
 
       - name: Run Claude PR Review
-        uses: anthropics/claude-code-action@3f854a8fb5146b39d5cbf8b57f70d80810e1366f # v1.0.198
+        uses: anthropics/claude-code-action@1f291e1cfe0f5fc21db2aef19af844591600ade7 # v1.0.206
         timeout-minutes: 60
         with:
           anthropic_api_key: ${{ secrets.ENG_ANTHROPIC_API_KEY }}
--- a/.github/workflows/typespec.yml
+++ b/.github/workflows/typespec.yml
@@ -29,7 +29,7 @@ jobs:
       - name : Setup Node.js environment
         uses : actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with :
-          node-version : '24.19.0'
+          node-version : '24.20.0'
 
       - name : Install tsp
         run : npm install -g @typespec/compiler@0.64.0
--- a/build.gradle
+++ b/build.gradle
@@ -194,6 +194,9 @@ test {
     environment 'QUAY_PAT', project.findProperty('QUAY_PAT') ?: environment['QUAY_PAT']
 
     systemProperty 'logback.configurationFile', 'src/test/resources/logback-test.xml'
+    // enable Basic authentication for HTTPS proxy tunnelling (CONNECT requests),
+    // required by `HttpClientFactoryProxyTest` - see also `HttpClientConfig.enableProxyAuthSchemes`
+    systemProperty 'jdk.http.auth.tunneling.disabledSchemes', ''
 
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -49,6 +49,7 @@ dependencies {
     implementation 'jakarta.persistence:jakarta.persistence-api:3.0.0'
     implementation 'io.seqera:lib-cache-tiered-redis:1.1.0'
     implementation 'io.seqera:lib-lang:1.1.0'
+    implementation 'io.seqera:lib-httpx:2.6.0'
     implementation 'io.seqera:lib-serde:1.2.0'
     implementation 'io.seqera:lib-serde-moshi:1.1.0'
     implementation 'io.seqera:lib-mail:1.3.1'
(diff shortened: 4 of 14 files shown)


ewels:
Note that another approach here could be to pull in libseqera to do more of the work and avoid having a duplicate fix to the other linked PRs. As that's a bigger change with wider architectural decisions to make I will leave that alone for now though.

ewels:
Following up on the comment above: this PR has now been reworked to do exactly that — it pulls in the shared libseqera implementation instead of carrying a duplicate fix.

- The branch was rebased on current `master` (resolving the conflicts from the docs restructure and the kebab-case config key rename).
- The bespoke `ProxySelector`/`Authenticator`/no-proxy matching has been deleted and replaced with **`io.seqera:lib-httpx:2.6.0`** (`HxProxyConfig`) — the same library and version-line used to fix this in Nextflow via nextflow-io/nextflow#7331 — so Wave and Nextflow now share identical proxy semantics.
- What remains Wave-side is only the resolution layer: the `wave.httpclient.proxy.*` settings with `HTTPS_PROXY`/`HTTP_PROXY`/`NO_PROXY` env-var fallback (`ProxyConfig`, modelled on Nextflow's `nextflow.util.ProxyConfig`), the `HttpClientFactory` wiring, and the `disabledSchemes` startup defaulting.

One behavioural note from adopting the shared library: no-proxy entries follow lib-httpx semantics (bare hostnames also match sub-domains, `.domain`/`*.domain` match sub-domains only, loopback always bypasses; CIDR entries are not supported). The PR description has been updated to reflect all of this.

---
_Generated by [Claude Code](https://claude.ai/code)_

pditommaso:
Pushed a follow-up commit that reworks this to consume the shared **`io.seqera.util.net.ProxyConfig`** (from `io.seqera:lib-util-net:0.2.0`, now published) instead of Wave carrying its own resolver — taking the change from +924/−20 down to **+109/−7** vs master.

What changed on top of the original commit:
- **Removed** `io.seqera.wave.http.ProxyConfig` (166 lines) and the `FakeProxyServer` test double — the URI/env parsing, no-proxy matching and proxy authenticator now live in lib-util-net (shared with Nextflow), where they're unit- and integration-tested.
- **Dropped the `lib-httpx` dependency** from Wave (it was only used for `HxProxyConfig`, which lib-httpx 2.7.0 also removes in favour of the same `ProxyConfig`).
- `HttpClientConfig` resolves the proxy from the `wave.httpclient.proxy.*` settings (scoped to Wave's clients) or, from the environment, via `ProxyConfig.setupFromEnvironment(System.getenv())` — which installs it JVM-wide (system properties + default authenticator), mirroring Nextflow so `HTTPS_PROXY`/`NO_PROXY` behave identically across both.
- Removed the redundant `jdk.http.auth.proxying.disabledSchemes` mutation (only the `tunneling` one is needed).

The resolver work landed in seqeralabs/libseqera#122 (lib-util-net) and #123 (lib-httpx 2.7.0), both merged and published. Compiles against the published artifacts; the proxy behaviour is covered by lib-util-net's tests plus lib-httpx's end-to-end proxy-auth integration test.

The PR description above still reflects the original approach — worth refreshing to match the surgical version.

pditommaso:
Superseded by #1125, which I'd propose closing this in favour of.

Same goal (nextflow-io/nextflow#7305), but #1125 consumes the shared `io.seqera.util.net.ProxyConfig` from `io.seqera:lib-util-net:0.2.0` rather than a Wave-local resolver — so Wave carries no `ProxyConfig`/`FakeProxyServer` of its own and drops the `lib-httpx` dependency, taking the change from +924/−20 down to +109/−7. The proxy URI/env parsing, no-proxy matching and authenticator now live in one place shared with Nextflow (nextflow-io/nextflow#7587) and lib-httpx, where they're unit- and integration-tested.

Thanks @ewels — this PR is what drove extracting the resolver into the shared library.

ewels:
Perfect, thanks!

PR #1080, opened 2026-07-13
chore: pin postgres containers + bump test containers

gavinelder:
The following bumps test containers and the Postgres version.


pditommaso (approved):
LGTM. Correct Testcontainers 2.0 module rename (`testcontainers-postgresql`) and BOM alignment; postgres pinned consistently to 16.14 across compose + TC JDBC URLs. Main risk was the 1.x→2.0 major bump affecting all GenericContainer-based test helpers — confirmed green CI clears it. Nice cleanup.

PR #1083, opened 2026-07-15
fix: install jq for R (bioconda) post-link scripts in conda v2 builds

pditommaso:
## Problem

Follow-up to #1070. Conda containers with R/bioconductor **data** packages (e.g. `bioconductor-tximeta` -> `GenomeInfoDbData`) still failed at runtime with:

```
Error: package or namespace load failed for 'GenomeInfoDb':
 there is no package called 'GenomeInfoDbData'
```

even after the `which` fix from #1070.

## Root cause

Bioconductor data packages install their payload via the conda post-link script `installBiocDataPackage.sh`, which parses the download URL with `yq`. `yq` shells out to **`jq`**, and the `micromamba:2-amazon2023` build image ships neither `which` (fixed in #1070) nor `jq`.

Without `jq`, `yq` errors out, the resolved filename comes back empty, the tarball is never downloaded, and the data package is **silently** not installed. The post-link failure is non-fatal to `micromamba install`, so the build still produces a broken image.

## Reproduced

Real Docker build on `mambaorg/micromamba:2-amazon2023`:

- **Without jq:** `yq: Error starting jq: ... No such file or directory: 'jq'` -> `FN=` empty -> R library missing `GenomeInfoDbData`.
- **With `conda-forge::jq`:** `yq` resolves the filename, `curl` downloads `GenomeInfoDbData_1.2.15.tar.gz` from bioconductor.org, `R CMD INSTALL` succeeds, package present.

Also confirmed that switching the base image to plain `mambaorg/micromamba:2` (Debian) does **not** help — it ships `which` but still lacks `jq`, so the same failure occurs. `jq` is required regardless of base image.

## Fix

Install `conda-forge::jq` alongside `conda-forge::which` in the build stage across all four `conda-micromamba-v2` templates and both `conda-pixi-v1` templates. Template assertion tests updated.

### DIFF
--- a/src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-file.txt
+++ b/src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-file.txt
@@ -1,8 +1,8 @@
 FROM {{mamba_image}} AS build
 USER root
 COPY --chown=$MAMBA_USER:$MAMBA_USER conda.yml /tmp/conda.yml
-# expose `which` at /usr/bin/which for R (bioconda) post-link scripts; the amazon2023 base image lacks it
-RUN micromamba install -y -n base conda-forge::which \
+# expose `which`+`jq` for R (bioconda) post-link scripts (installBiocDataPackage.sh); the amazon2023 base image lacks them
+RUN micromamba install -y -n base conda-forge::which conda-forge::jq \
     && ln -sf "$MAMBA_ROOT_PREFIX/bin/which" /usr/bin/which \
     && (micromamba install -y -n base -f /tmp/conda.yml > /tmp/mamba.log 2>&1 \
     && cat /tmp/mamba.log \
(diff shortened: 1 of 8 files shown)


pditommaso:
Closing in favour of fixing this at the correct layers.

**Root cause (upstream):** the conda-forge `yq` package does not declare `jq` as a runtime dependency, even though it is a jq wrapper. bioconductor data-package post-link scripts (`installBiocDataPackage.sh`) call `yq`, which fails without `jq`, so `GenomeInfoDbData` is silently not installed.
- Upstream issue: https://github.com/conda-forge/yq-feedstock/issues/41

**Interim fix (module level):** pin `jq` in the affected nf-core module's `environment.yml`, which — given the `which` fix already merged in #1070 — is sufficient to make the build succeed (verified with a real `micromamba:2-amazon2023` build).
- nf-core module patch: https://github.com/nf-core/modules/pull/12362

Given the durable fix is upstream and the immediate unblock is at the module level, patching every Wave conda build to install `jq` is unnecessary. Closing this PR.

PR #1087, opened 2026-07-16
Bugfix: do not serve freeze/mirror images via the Wave proxy token path

gavinelder:
## Summary

Bug fix — ensure freeze (and mirror) images are not pulled via Wave.

Freeze and mirror requests publish the image to the caller's target registry and return a **direct** registry reference; those images are meant to be pulled directly from that registry. However the ephemeral `/v2/wt/<token>/` proxy route still resolved and served them for the token's lifetime, leaving Wave in the pull path. This change rejects durable (freeze/mirror) requests on the proxy token path, so a frozen/mirrored image can only be pulled directly from the target registry.

## Changes

- `RouteHandler.parse`: when a `wt/<token>` request resolves to a `durable()` request (freeze/mirror), return `404` instead of routing the proxy pull. Status/detail lookups by request id are unaffected (they do not go through this path).
- `RouteHandlerTest`: add coverage asserting freeze and mirror proxy pulls are rejected; existing non-durable resolution cases are unchanged.

## Testing

- `RouteHandlerTest` passes 32/32, including the new freeze/mirror rejection cases and the existing non-durable (build/augment/proxy) resolution cases — no regression.

### DIFF
--- a/src/main/groovy/io/seqera/wave/core/RouteHandler.groovy
+++ b/src/main/groovy/io/seqera/wave/core/RouteHandler.groovy
@@ -62,6 +62,12 @@ class RouteHandler {
             if( !request ) {
                 throw new NotFoundException("Unknown Wave container for token '$token'")
             }
+            // freeze and mirror requests publish the image to the target registry and are
+            // meant to be pulled directly from there - they must not be served via the Wave
+            // proxy token path (see also ContainerRequest.durable)
+            if( request.durable() ) {
+                throw new NotFoundException("Wave container for token '$token' is not available via the proxy")
+            }
             // the image name (without tag) must match
             final coords = request.coordinates()
             if( image != coords.image )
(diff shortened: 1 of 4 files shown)


cristianrcv (approved):

pditommaso (approved):
Looks good — solid fix. 🙌

I pushed one small follow-up on top: made the rejection message actionable so it points the caller at the direct image reference (`must be pulled directly from '<image>'`) rather than a generic "not available via the proxy", and tightened the test to assert the message carries that reference. Kept the 404 as-is — it's the right registry-conventional code here.

One thing worth a quick sanity check (non-blocking, happy to take offline): the guard is unconditional, but `makeResponseV1` still hands V1 (`/container-token`) freeze clients a `wt/<token>` URL, whereas `makeResponseV2` already returns the direct image. If legacy V1 freeze is still in use anywhere, those pulls would now 404. If V1 freeze is effectively dead, all good — just flagging.

pditommaso (changes requested):
Actually the V1 freeze path should be investigate more before merging.

pditommaso (approved):
Sorted

PR #1090, opened 2026-07-17
Return direct image for durable (freeze/mirror) requests on the V1 endpoint

pditommaso:
## Summary

Stacked on top of #1087.

#1087 makes the `/v2/wt/<token>` proxy path reject durable (freeze/mirror) requests with a 404 — those images are meant to be pulled directly from the target registry. But the two response builders were asymmetric:

- `makeResponseV2` already returns the **direct** target-registry image for durable requests (`target = data.durable() ? data.containerImage : waveImage`), so V2 clients never receive a `wt/<token>` freeze url.
- `makeResponseV1` returned `targetImage = waveImage` **unconditionally**, so a V1 (`/container-token`) freeze client is handed the proxy url — which #1087 now 404s.

Freeze is **not** v2-gated (only mirror is), and `/container-token` is still live (`@Deprecated`), so a pre-`24.04` Nextflow client (or any other V1 freeze client) would break.

## Impact assessment

- **Current Nextflow: unaffected.** The Wave client is V2-only since #4906 (Nextflow v24.04.0, Apr 2024) and resolves the run image from `targetImage`, which V2 already returns as the direct image.
- **Pre-24.04 Nextflow / other V1 freeze clients:** would 404 on the proxy pull after #1087. This change fixes them — they now receive the direct image in `targetImage` and pull it directly (the correct freeze behavior).

## Changes

- `ContainerHelper.makeResponseV1`: resolve `targetImage` to the direct image (`data.containerImage`) for `durable()` requests, matching `makeResponseV2`. The token is intentionally retained in the V1 response to avoid surprising legacy parsers — only `targetImage` changes.
- `ContainerHelperTest`: add V1 freeze + mirror cases asserting the direct image is returned.

### DIFF
--- a/src/main/groovy/io/seqera/wave/util/ContainerHelper.groovy
+++ b/src/main/groovy/io/seqera/wave/util/ContainerHelper.groovy
@@ -116,7 +116,9 @@ class ContainerHelper {
     }
 
     static SubmitContainerTokenResponse makeResponseV1(ContainerRequest data, TokenData token, String waveImage) {
-        final target = waveImage
+        // freeze/mirror images are published to the target registry and must be pulled
+        // directly from there - the Wave proxy token path does not serve them (see RouteHandler)
+        final target = data.durable() ? data.containerImage : waveImage
         final build = data.buildNew ? data.buildId : null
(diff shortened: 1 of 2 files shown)


gavinelder (approved):
Thank you I was weighing up this approach and was unsure how best to proceed.

PR #1106, opened 2026-08-11
docs: Refresh self install

christopher-hakkaart:


christopher-hakkaart:
@copilot resolve the merge conflicts in this pull request

pditommaso (commented):
## Review: self-install docs refresh

The restructure is the right shape — two-path index, assemblable manifests, a shared verification page, settings pushed out to `reference.md`. But the pages don't run yet. Reviewed at `1e4e10d`; every item below re-checked by hand against `src/main`.

### Won't work

Two that stop a deployment cold:

1. **`aws-build.md:377` enables scan without `wave.scan.reports.path`.** `ObjectStorageOperationsFactory.create0` throws on a null path, and `Application.groovy:43` starts Micronaut with `eagerInitSingletons(true)` — so Wave won't boot. `reference.md:378` already marks the key required.
2. **`aws-build.md:272` creates `wave-build-pvc` in namespace `wave`**, but line 362 runs build jobs in `wave-build`. PVCs are namespace-scoped, so every build/scan/mirror pod stays Pending.

And a regression: **the merge commit reverted six keys in `reference.md` from kebab-case back to camelCase**, undoing #1096 (master had all six right). They still bind from YAML because Micronaut hyphenates every key it loads, but `WAVE_BLOBCACHE_STORAGE_ACCESSKEY` no longer resolves — and lines 563/581 still say `blob-cache.*`, so the section contradicts itself.

Four more that fail silently:

- `DOCKER_USER`/`DOCKER_PAT` in `wave.env` (`docker-compose.md:110`) do nothing. Nothing in `src/main` reads those names — they work only via the `config.yml` the image ships (`build.gradle:184`), which this guide tells you to mount over.
- The `MICRONAUT_ENVIRONMENTS` strings in `configure-wave.md:78,130,326` re-add `lite`, and `WaveLite` gates out `JobManager`, `JobServiceImpl`, both job queues and cleanup via `@Requires(missingBeans = WaveLite)`.
- The `endpoints.health` guard master shipped (`health.jdbc` / `health.disk-space` off) was dropped, while `/health` is now the liveness probe in both guides. A brief Postgres blip restarts every replica.
- `WAVE_JVM_OPTS` was dropped from a 4 GiB pod; `src/main/jib/launch.sh` defaults to `-Xmx850m`.

Also: the registry snippet at `kubernetes-lite.md:146` is a second top-level `wave:` block (Wave loads one YAML document); that page never says `kubectl apply`; its ingress has no `ingressClassName` or annotations so `/service-info` can't pass; the `<s3-bucket>` the ConfigMap depends on is never created; `noarch` is missing from the node selector so mirror pods skip the build node group; the production checklist uses the legacy `wave.allowAnonymous` alias instead of `wave.capabilities.anonymous-access`; and `how-wave-works.md:126` still links `./configuration.md`.

### Simplify

Length is concentrated in three blocks: the seven-registry pre-creation matrix (`aws-build.md:59-89` — on an EKS page), the IAM CLI narration (109-132, replaceable with one AWS link), and the SMTP table (`configure-wave.md:99-109`). Beyond that: Bottlerocket and "Production enhancements" belong in troubleshoot / the existing checklist, IRSA is explained three times across two files, and `configure-wave.md` documents an EC2 instance profile for Docker-based builds that `index.md:23` says aren't supported.

A stacked PR against this branch follows with the fixes.

justinegeffen (approved):

christopher-hakkaart:
Thanks @pditommaso - changes are merged with a lite deslop/style guide

docs/install/post-install.md
```diff
@@ -0,0 +1,106 @@
...
+
+## If a check fails
+
+If a check does not pass, match the symptom to one of these common causes:
+
+- **Augmentation or build fails with an authentication error**: Wave cannot authenticate to the target registry. See [Registry push and authentication failures](../troubleshoot.md#registry-push-and-authentication-failures).
+- **Build fails partway through a push**: The target repository may not exist or may lack push scope. See [Registry pre-creation](reference.md#registry-pre-creation).
+- **Freeze is rejected even with a build repository set**: The repository sits inside an operator-reserved prefix. See [Freeze and user-supplied build repositories](aws-build.md#freeze-and-user-supplied-build-repositories).
+
+When the checks pass, continue to the [production checklist](configure-wave.md#production-checklist) to prepare the deployment for production.
```
christopher-hakkaart, line 106:
```suggestion
When the checks pass, continue to the [production checklist](configure-wave.md#production-checklist) to prepare the deployment for production.

```

PR #1114, opened 2026-08-19
Add gzip to image templates

### LINKED ISSUE #1113: `gzip` missing from singularity base image
Hi,

Would it be possible to add `gzip` by default to the singularity base image?
nf-core is migrating to wave containers where possible, but CI tests using the common `echo "" | gzip > test.gz` pattern for stub tests fail using the generated singularity images, since `gzip` is not available.

I'm aware it's possible to add `gzip` as a dependency to each module as they are updated, but I think it would be cleaner to just add it to the base image as a common utility, so the singularity images are on par with the docker ones

Thanks
Matthias

matthdsm:
Fixes #1113

### DIFF
--- a/src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-packages.txt
+++ b/src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-packages.txt
@@ -3,7 +3,7 @@ USER root
 # expose `which` at /usr/bin/which for R (bioconda) post-link scripts; the amazon2023 base image lacks it
 RUN \
     micromamba install -y -n base conda-forge::which \
-    && ln -sf "$MAMBA_ROOT_PREFIX/bin/which" /usr/bin/which \
+    && ln -sf "$MAMBA_ROOT_PREFIX/bin/which" /usr/bin/which conda-forge::gzip\
     && (micromamba install -y -n base {{channel_opts}} {{target}} > /tmp/mamba.log 2>&1 \
     && cat /tmp/mamba.log \
     || (cat /tmp/mamba.log >&2 && grep -q __cuda /tmp/mamba.log \
--- a/src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-packages.txt
+++ b/src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-packages.txt
@@ -2,7 +2,7 @@ FROM {{mamba_image}} AS build
 USER root
 # expose `which` at /usr/bin/which for R (bioconda) post-link scripts; the amazon2023 base image lacks it
 RUN \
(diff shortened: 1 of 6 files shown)


src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-packages.txt
```diff
@@ -3,7 +3,7 @@ USER root
 # expose `which` at /usr/bin/which for R (bioconda) post-link scripts; the amazon2023 base image lacks it
 RUN \
     micromamba install -y -n base conda-forge::which \
-    && ln -sf "$MAMBA_ROOT_PREFIX/bin/which" /usr/bin/which \
+    && ln -sf "$MAMBA_ROOT_PREFIX/bin/which" /usr/bin/which conda-forge::gzip\
```
ewels: This looks wrong

src/main/resources/templates/conda-pixi-v1/dockerfile-conda-file.txt
```diff
@@ -4,7 +4,7 @@ COPY conda.yml /opt/wave/conda.yml
 WORKDIR /opt/wave
 
 RUN pixi init --import /opt/wave/conda.yml \
-    && pixi add conda-forge::which \
+    && pixi add conda-forge::which conda-forge::gzip\
```
ewels:
```suggestion
    && pixi add conda-forge::which conda-forge::gzip \
```

ewels (dismissed):

src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-packages.txt
matthdsm: fixed!

matthdsm:
To reduce the load on the cache, I could also limit the changes to only the singularity files, since gzip is already available in the docker images.
I added it to both for now to have parity in the `gzip` versions between the containers

src/main/resources/templates/conda-micromamba-v2/dockerfile-conda-packages.txt
```diff
@@ -2,7 +2,7 @@ FROM {{mamba_image}} AS build
 USER root
 # expose `which` at /usr/bin/which for R (bioconda) post-link scripts; the amazon2023 base image lacks it
 RUN \
-    micromamba install -y -n base conda-forge::which \
+    micromamba install -y -n base conda-forge::which conda-forge::gzip\
```
ewels:
```suggestion
    micromamba install -y -n base conda-forge::which conda-forge::gzip \
```

pditommaso:
Closing because it can be managed via nextflow config as mentioned in the issue https://github.com/seqeralabs/wave/issues/1113#issuecomment-5339731764

PR #1115, opened 2026-08-19
Replace lib-data-stream-redis with lib-data-workqueue

pditommaso:
Moves the job queues from the message stream API onto the work queue API of libseqera: `lib-data-stream-redis:1.2.0` → `lib-data-workqueue:2.0.0` + `lib-data-workqueue-redis:2.0.0`.

`lib-data-workqueue-redis` is `lib-data-stream-redis` 1.5.0 re-landed on lease semantics — identical Redis stream layout (`xadd`/`xreadGroup`/`xautoclaim`, `xack`+`xdel` on settle, same `data` field). Queue keys (`jobs-pending/v2`, `jobs-queue/v1`) and the consumer group are left untouched, so **no migration is needed** and messages queued before the upgrade are still delivered afterwards.

### Changes

| | |
|---|---|
| `build.gradle` | swap the stream artifact for the two work-queue artifacts |
| `service/data/workqueue/BaseWorkQueue` | was `BaseMessageStream`, now on `AbstractWorkQueue`; same Moshi encoding strategy |
| `service/data/workqueue/RedisWorkQueueConfigBean` | was `RedisStreamConfigBean`, now implements `RedisWorkQueueConfig` |
| `JobPendingQueue` / `JobProcessingQueue` | extend `BaseWorkQueue`; queue ids unchanged |
| `JobManager` | `(job) -> boolean` becomes `(job, lease) -> Decision.ACK/RETRY` |
| `docs/configuration.md` | "Message stream" section becomes "Work queue" |

`launchJob`/`processJob` are unchanged and still return a boolean, mapped onto `ACK`/`RETRY` at the `addConsumer` site — so `JobManager` keeps its own admission control and the `maxRunningJobs` cap is unaffected.

### Why: work queue vs message stream

| | message stream (before) | work queue (after) |
|---|---|---|
| **Slow consumer** | An entry's idle clock only resets on delivery/claim, so a job outlasting `claim-timeout` can be claimed and run **concurrently by another replica**. The timeout had to exceed worst-case processing time — hence 45s in prod. | A lease is renewed at `visibility-timeout/4` while the owner is alive (one `XCLAIM JUSTID` per queue per tick), holding the idle clock near zero. The timeout now only governs recovery from a **dead** owner. |
| **Shutdown** | `thread.interrupt()` then `join(1s)` — severs a consumer mid-processing, and an interrupt can hand a RESP-desynced connection back to the Jedis pool. | Cooperative: a `closing` flag checked at loop head, waits (default 10s) for the current cycle, **never interrupts**. A `processJob` cycle completes its `notifyJobCompletion`/`cleanup` before teardown. `awaitQuiescent(timeout)` available for a bounded drain. |
| **Retry pacing** | Any delivery counted as progress, so a cycle that only returned `false` skipped the poll-interval sleep. | Only `ACK`/`DEFERRED` count as progress, so a queue that is merely retrying (job still running, pending queue full) paces at its poll interval. |
| **Ownership safety** | — | Renewal does an `XPENDING` ownership check first: an entry that drifted to another consumer is dropped and logged rather than seized back — the residual duplicate window is observable, not silent. |

Not used by this PR, but now available: `MessageConsumer.ready()` (an admission gate checked *before* claiming — the natural home for the `maxRunningJobs` check, which currently claims a message and then refuses it), `MessageLease`/`DEFERRED` with `retryAfter(delay)`, and lease/renewal/saturation metrics. Still **at-least-once**: renewal narrows the duplicate window, it does not close it.

### Rollout

No Redis migration. Entries queued by the old code are read as-is (the integration test logs `consume group=wave-message-stream already exists`), and entries left pending by a terminating pod are reclaimed after the visibility timeout — in-flight jobs resume. Mixed-version replicas are safe both ways: an entry held by a new replica is renewed so an old replica's 45s-min-idle claim never sees it; an entry held by an old replica may be claimed after 45s idle, exactly as today.

Config keys are renamed `wave.message-stream.*` → `wave.work-queue.*` (`claim-timeout` → `visibility-timeout`), so **the matching rename in platform-deployment (prod + stage, `45s`) must land with this**. That change is prepared but not yet pushed.

:warning: The one hazard is **rollback**: revert the image with the new ConfigMap in place and the old code finds no `message-stream.claim-timeout`, falling back to a 5s claim timeout with no renewal. Keeping both key sets in the ConfigMap for one release makes it order-independent in both directions. Running pods are unaffected either way — the ConfigMap is mounted via `subPath` (never updated in place) and the Deployment has no config checksum annotation, so config is only read at container start.

Also note the cooperative `close()` can take up to 10s per queue, sequentially across two beans, against the default 30s `terminationGracePeriodSeconds`. In practice the dispatcher exits at its next loop-head check unless mid-consume; raising the grace period would add headroom.

The consumer group keeps defaulting to `wave-message-stream` deliberately — it is a wire identifier now, and renaming it would strand the pending-entry list.

### Testing

- `compileGroovy` clean
- `io.seqera.wave.service.job.*` — 39 tests, 0 failures
- `BuildStoreRedisTest` (7) and `RegistryControllerRedisTest` (2) green against real Redis; logs confirm `RedisWorkQueue` reusing the existing consumer group and picking up the config bean (`lease renewal period=1250ms`)

Full suite not run locally — left to CI.


pditommaso:
@claude review

pditommaso:
Verified the three open questions against the `lib-data-workqueue:2.0.0` source directly:

1. **`length()` counts leased-but-unsettled entries** — it's `XLEN`, and an entry only leaves the stream on ACK (`xack`+`xdel` together). Running/RETRY'd entries stay counted, so the `maxRunningJobs` cap still works. Identical to the old stream.
2. **No delivery-count cap / dead-lettering** — RETRY does no Redis call, there's no attempt counter, and the only pruning (age backstop) is liveness-gated and just re-recovers the entry rather than dropping it. wave uses synchronous ACK/RETRY (no DEFERRED/liveness), so leases settle within the `consume()` call and nothing accumulates. A long build is not stranded; still at-least-once.
3. **RETRY re-visibility is the visibility timeout (45s in prod), not ~1s** — but the old stream behaved identically (claim-timeout was also 45s), so completion-detection latency is unchanged. The retry-pacing row is about the dispatcher loop no longer busy-spinning, not per-entry redelivery timing.

So all three are as the description says — no behaviour change from the swap.

Separately actioning the two nits (mapping-site test coverage; PR body should read `docs/install/reference.md`, not `docs/configuration.md`).

pditommaso:
@munishchouhan can you please have a review at this PR ?

munishchouhan:
### `bindLiveness` is never called, so the lease age backstop prunes every lease

`JobManager.init()` accepts the `lease` and discards it:

```groovy
pendingQueue.addConsumer((job, lease)-> launchJob(job) ? Decision.ACK : Decision.RETRY)
processingQueue.addConsumer((job, lease)-> processJob(job) ? Decision.ACK : Decision.RETRY)
```

With no probe bound, `Lease.isOwnerAlive()` returns `false` unconditionally, so the backstop in `RedisWorkQueue:463` collapses to a plain age check:

```java
if (!lease.isHeldForRelease() && lease.ageNanos() > maxLeaseAgeNanos && !lease.isOwnerAlive())
```

`MessageLease#bindLiveness` states it directly: *"Without a bound probe the age backstop applies unconditionally."*

**Effect:** a consumer cycle outliving `maxLeaseAge` (3 × visibility timeout → 15s on the 5s default) stops being renewed; one visibility timeout later another replica can claim the entry while `launchJob`/`processJob` is still running. Better than the 45s pre-PR window, but the timeout still governs *slow* owners, not only dead ones — which is the claim in the description.

**Fix:**

```groovy
pendingQueue.addConsumer((job, lease)-> {
    lease.bindLiveness(() -> true)
    launchJob(job) ? Decision.ACK : Decision.RETRY
})
```

`() -> true` is accurate for a synchronous consumer: the lease is registered immediately before `accept()` and unregistered on every return path, so a still-registered lease means the handler is still running.

Trade-off to confirm before applying: this renews for as long as the handler runs, so a permanently hung handler would hold the entry indefinitely instead of being recovered after `maxLeaseAge`. Fine provided the k8s/Docker calls in `jobService.status()` and `dispatcher.launchJob()` are timeout-bounded.

pditommaso:
@munishchouhan thanks for the detailed analysis — the mechanics are exactly right, but after digging into the behaviour I've decided **not** to bind a liveness probe here, and to keep the lease discarded. Rationale:

**1. Discarding the lease preserves the previous (message-stream) behaviour.** On `master`, a consumer that runs past the `claim-timeout` (45s in prod) has its message re-claimed by another consumer via `xautoclaim` — a slow/stalled owner loses the entry and it is recovered (with a possible duplicate). The work-queue age backstop (`3 × visibility-timeout`) reproduces that same class of behaviour: a slow owner's lease is released and the entry recovered. So the current code is a faithful migration, not a regression.

**2. `bindLiveness(() -> true)` would actually diverge from `master`.** With a constant-`true` probe the backstop never fires, so a *stalled* consumer holds the entry forever with no takeover. And this isn't hypothetical: the k8s client is built with `readTimeout = 0` (infinite) — `ClientBuilder.cluster()/kubeconfig().build()` never overrides it — so a stalled `logsPod` read inside `processJob` can block indefinitely. `max-duration` can't rescue it either, because it's checked *after* `status()`/`logsPod` returns. So `() -> true` would trade "possible duplicate on a slow poll" (which `master` also has) for "job wedged forever on a stalled read" (which `master` does **not** — it re-claims). Net worse.

**3. The right way to suppress duplicates is a bounded k8s read timeout, not a constant probe.** With a finite read timeout a stalled read errors, `logsPod` returns `null`, the poll returns and recovers — and only *then* is `() -> true` safe. That belongs with the async model, not this PR.

**Follow-up:** #1129 tracks moving `JobManager` to the `DEFERRED`/async consumer model (as `lib-cmd-queue-redis` does in sched), where `bindLiveness(() -> !task.isDone())` **is** the correct call and we also get parallel status polling — plus the prerequisite k8s read timeout.

Separately I've made the config forward-compatible: the `wave.work-queue.*` settings fall back to the legacy `wave.message-stream.*` keys and default to the current prod values (`45s`), so prod/stage keep their `45s` with no deployment change. Docs and the `lib-data-workqueue` README (sync vs async `bindLiveness` guidance) updated accordingly.

pditommaso:
Pushed b5fc2b5f5: the `wave.work-queue.*` settings now fall back to the legacy `wave.message-stream.*` keys and default to the current prod values (`45s`), so prod/stage keep their behaviour with no deployment change. Configuration reference updated to match.

Consumer model is kept synchronous (lease intentionally discarded) as explained above; the async `DEFERRED` model + `bindLiveness(() -> !task.isDone())` is tracked in #1129.

munishchouhan:
pushed fix for:
```
Message: Failed to inject value for parameter [visibilityTimeout] of method [setVisibilityTimeout] of class: io.seqera.wave.service.data.workqueue.RedisWorkQueueConfigBean

Message: Error resolving property value [${wave.work-queue.visibility-timeout:${wave.message-stream.claim-timeout:45s}}]. Unable to convert value [45s}] to target type [Duration] due to: Unparseable date format (45s}). Should either be a ISO-8601 duration or a round number followed by the unit type
```

PR #1128, opened 2026-09-07
fix: report K8s builds as succeeded on Kubernetes 1.36+ (COMP-2368)

munishchouhan:
Fixes [COMP-2368](https://seqera.atlassian.net/browse/COMP-2368).

## Problem

On a Kubernetes 1.36 or later cluster every container build is reported as failed even though the image is built and pushed successfully. The build status endpoint answers with a self-contradicting payload and Nextflow terminates the run:

```json
{"id":"bd-5d1d489defad90d5_1","status":"COMPLETED","duration":187.24,"succeeded":false}
```

Reported by a customer on self-hosted Wave 1.32.4. Users lose ~45 minutes per container and only one container gets past the gate per attempt, so a 16-image pipeline needs many attempts. `errorStrategy = 'retry'` and `'ignore'` have no effect, because Nextflow overrides every soft error strategy to `TERMINATE` for an unrecoverable Wave error.

## Cause

`io.kubernetes:client-java` was pinned at 24.0.0, which models Kubernetes 1.33. A newer cluster returns pod level resource fields that `V1PodStatus` does not define, so the generated `validateJsonElement` throws:

```
IllegalArgumentException: The field `allocatedResources` in the JSON string is not defined
in the `V1PodStatus` properties.
  at io.kubernetes.client.openapi.models.V1PodStatus.validateJsonElement(V1PodStatus.java:663)
```

Two pod level fields are involved and either one is enough to trip the client: `status.allocatedResources` and `status.resources`. Which one appears in the message depends on the build job's resource requests — with `wave.build.k8s.resources.requests` configured the pod is Burstable and `allocatedResources` is populated (the customer case); without them the pod is BestEffort, `allocatedResources` is omitted as empty and the validator trips on `resources` instead. 24.0.0 defines neither, so the outcome is identical.

The Job itself parses fine, so the job status is `Succeeded`. The exception is thrown by the *carrier pod* lookup that follows and escapes `K8sJobOperation.status`. From there:

1. `JobServiceImpl.status` catches `Throwable` and returns `JobState.unknown(t.message)`.
2. `JobManager.state0` holds that unknown for the grace interval, then converts it to `FAILED` with a **null** exit status.
3. The FAILED state is a completion, so `notifyJobCompletion` fires, the queue entry is consumed and the build is never retried — which is why no `_2` build record exists for any affected digest.

Note the exception never reaches the job manager's `catch(Throwable)`/`notifyJobException` path, so the reported exit status is `null` rather than `-1`. The same `catch(Throwable)` in `JobServiceImpl.status` is present in v1.32.4, so this is also the mechanism on the customer's version.

The API response reads as contradictory only because `status: COMPLETED` means `duration != null`, ie. "finished", while `succeeded` means `exitStatus == 0`. A single code path sets both halves.

### Affected cluster versions

Pod level `allocatedResources`/`resources` are absent from the release-1.33 and release-1.34 API specs and first appear in release-1.35 — but the kubelet only populates them when the `InPlacePodLevelResourcesVerticalScaling` feature gate is on (release-1.35 `pkg/kubelet/kubelet_pods.go`):

```go
if utilfeature.DefaultFeatureGate.Enabled(features.InPlacePodLevelResourcesVerticalScaling) {
    apiPodStatus.Resources = kl.convertToAPIPodLevelResourcesStatus(pod, oldPodStatus)
    apiPodStatus.AllocatedResources = resourcehelper.PodRequests(pod, opts)
}
```

and that gate is `Alpha, default false` in 1.35, `Beta, default true` in 1.36 (`pkg/features/kube_features.go`). Its dependencies are all default-on by 1.36: `InPlacePodVerticalScaling` GA in 1.35, `PodLevelResources` beta since 1.34, `NodeDeclaredFeatures` beta in 1.36.

| release | exposure with client 24.0.0 |
|---|---|
| 1.33 and below | not exposed — the fields do not exist |
| 1.34 | only for pods using extended resource claims — 1.34 adds `PodStatus.extendedResourceClaimStatus`, which 24.0.0 also does not define, but it is omitted from the JSON unless set, so BuildKit pods do not hit it |
| 1.35 | only if the operator enables `InPlacePodLevelResourcesVerticalScaling` (alpha, and it also requires `NodeDeclaredFeatures`) |
| **1.36 and later** | **every completed job**, because the gate is beta/on by default and the node fills the fields in for every scheduled pod |

That the fields are always populated from 1.36 is why this is 100% of completed jobs rather than intermittent. On the cluster version question for COMP-2368, the customer payload proves either >= 1.36 or 1.35 with that alpha gate explicitly enabled; `kubectl version` would pin the exact minor.

## Changes

**Upgrade `client-java` and `client-java-api-fluent` to 24.0.0 -> 27.0.0.** 26.0.1 is the minimum version that defines the pod level resource attributes (25.0.1 does not); 27.0.0 is the latest release. With the fields modelled, the carrier pod parses, the real container exit code is read and the build log is preserved — `state.stdout` comes from `logsPod(pod)`, which is why the logs come back with the bump.

27.0.0's `V1PodStatus` matches the release-1.36 spec exactly, in both directions, so the bump covers 1.35 and 1.36.

Per review this PR is deliberately the bump alone. An earlier revision added a guard around the carrier pod lookup in `K8sJobOperation.status` that fell back to the Job level status on a parse failure; it was dropped because on a future version skew it would turn a hard failure into a quiet success with no build logs, and because it left the `V1Job` parse in `getJobStatus` uncovered anyway. Resilience against future skew is tracked separately — see the notes below.

## Testing

No new unit tests: the change is a dependency version, and whether the client model accepts the pod level resource fields is the kubernetes client library's own behaviour rather than something Wave can assert without pinning the library. The existing suite covers the call sites that consume the parsed model.

The k8s and job packages are green on the bump — 84 tests across 7 classes (`K8sServiceImplTest` alone is 45 and exercises the client API broadly), 0 failures. The full suite has 53 pre-existing failures across 17 classes in an environment without Docker Hub, AWS and Quay credentials; a baseline run of those classes on `master` reproduces an identical per-class breakdown, and no failure stack anywhere mentions kubernetes, `V1Pod` or gson.

Note a downgrade below 26.0.1 would reintroduce the parse failure, and with no guard in place it would surface exactly as described in the Cause section above.

### Verified against real clusters

Local Wave (`run_local.sh`, k8s build strategy) driving builds through the wave CLI against colima clusters, same Containerfile every time, pushing to a dev repo:

| cluster | pod level fields | client 24.0.0 | client 27.0.0 |
|---|---|---|---|
| k3s v1.35.8, stock | absent | not reproducible — build succeeds and is reported succeeded | succeeded |
| k3s v1.35.8, alpha gates enabled | present | `Job=Complete`, image pushed, Wave reports `exit=null; status=FAILED`, CLI `succeeded:false` | `exit=0`, build log captured, CLI `succeeded:true` |
| k3s v1.36.4, stock, no gates | present | `Job=Complete`, image pushed, Wave reports `exit=null; status=FAILED`, CLI `succeeded:false` | `exit=0`, `{"status":"COMPLETED","succeeded":true}`, full build log |

In both failing runs the registry tag was pushed before Wave declared the build failed, which is the customer symptom exactly. The 27.0.0 runs were made on the revision that still carried the guard, but the guard never fired in either — no parse warning is present in either log — so the outcome is attributable to the bump alone. The boundary was cross-checked cluster-free by deserializing the real API payloads through `client-java-api` 24.0.0 / 25.0.1 / 26.0.1 / 27.0.0.

## Notes for review

- **Acceptance criterion 4 in the ticket is inaccurate.** It expects the digest lookup in `ContainerBuildServiceImpl.onJobCompletion` to stop a failed push reporting success. It does not — `BuildResult.succeeded()` is `done() && exitStatus == 0` and never reads `digest`. What actually prevents it is that a failed push makes BuildKit exit non-zero, so the pod exit code is non-zero and the build is reported failed. Making a null digest fail the build outright would be a behaviour change beyond this ticket and carries its own risk (a transient `getImageDigest` failure would turn good builds into failures), so it is left alone.
- **Resilience against a future version skew is not addressed here.** The preferred shape is the permissive-deserialisation option rather than a per-call-site guard: rebuild the Gson from `JSON.createGson()` plus the five public value adapters, which drops the strict per-model unknown-field check across every model at once (measured: that parses the failing payload on an unpatched 24.0.0 client). Note `JSON.setLenientOnJson` does not help, as the check is a hard-coded throw in the generated model. Two call sites would benefit: `getLatestPodForJob` and the `V1Job` parse in `getJobStatus`, which embeds a full `V1PodTemplateSpec`.
- **Not covered here:** verifying scan and mirror job status on a real 1.36 cluster.

[COMP-2368]: https://seqera.atlassian.net/browse/COMP-2368?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


robsyme:
Reviewed this against the customer payload and checked the two version claims upstream. The guard looks right to me, including catching `IllegalArgumentException` only. Wanted to drop you a line about one potential correction, and a simplification.

From our friend Claude:

## The trigger is Kubernetes 1.35+, not 1.34

Pod-level `PodStatus.allocatedResources` (and `PodStatus.resources`) are absent from release-1.33 and release-1.34, and first appear in release-1.35:

```
for rel in 1.33 1.34 1.35 1.36; do
  curl -sf https://raw.githubusercontent.com/kubernetes/kubernetes/release-$rel/api/openapi-spec/v3/api__v1_openapi.json \
  | python3 -c 'import json,sys; p=json.load(sys.stdin)["components"]["schemas"]["io.k8s.api.core.v1.PodStatus"]["properties"]; print("allocatedResources" in p)'
done
1.33  False
1.34  False
1.35  True
1.36  True
```

1.34 does add fields that 24.0.0 has never heard of, `PodStatus.extendedResourceClaimStatus`, `PodSpec.hostnameOverride` and `Container.restartPolicyRules`, but every one of them is omitted from the JSON unless something sets it, so none of them ever fires. `allocatedResources` is the one that is always present, because the node fills it in for every scheduled pod:

> AllocatedResources is the total requests allocated for this pod by the node. If pod-level requests are not set, this will be the total requests aggregated across containers in the pod.

That is also why this is 100% of completed jobs rather than an intermittent thing. Worth changing in the PR description and in the `K8sJobOperation` code comment (`introduced by Kubernetes 1.34`), and in the issue title and Scope section on COMP-2368, because it decides who else is exposed: clusters at 1.34 and below are not.

It also closes the open question about AbbVie's cluster version. The payload in the ticket is proof of >= 1.35 on its own; `kubectl version` would only pin the exact minor.

On headroom, 27.0.0 is a good pick beyond just being latest: its `V1PodStatus` matches release-1.36 in full, including `nodeAllocatableResourceClaimStatuses`, so the bump covers both 1.35 and 1.36. (1.36's `PodSpec.schedulingGroup` is unknown to 27.0.0, but is again omit-when-unset.)

## The permissive-deserialisation follow-up needs no custom TypeAdapterFactory

All 694 strict per-model `CustomTypeAdapterFactory` instances are registered in exactly one place, the static initialiser of `io.kubernetes.client.openapi.JSON`. `JSON.createGson()` itself registers none of them. So rebuilding the Gson from `createGson()` plus the five public value adapters yields a client that ignores unknown keys, because Gson's reflective adapter skips them by default:

```java
JSON.setGson(JSON.createGson()
        .registerTypeAdapter(Date.class, new JSON.DateTypeAdapter())
        .registerTypeAdapter(java.sql.Date.class, new JSON.SqlDateTypeAdapter())
        .registerTypeAdapter(OffsetDateTime.class, new JSON.OffsetDateTimeTypeAdapter())
        .registerTypeAdapter(LocalDate.class, new JSON.LocalDateTypeAdapter())
        .registerTypeAdapter(byte[].class, new JSON.ByteArrayAdapter())
        .create());
```

Measured: that parses the failing payload on the **unpatched 24.0.0** client, exit code and timestamps included. The gson-fire `V1Status` preprocessor and `V1MetadataExclusionStrategy` survive, because they come from `createGson()`. The trade is losing required-field and array-shape validation across all models, which looks acceptable for read-only consumption of pod and job status, and it removes the class of failure instead of moving the ceiling up two minors. Your point that `JSON.setLenientOnJson` does not help is confirmed.

## How this was checked

Cluster-free reproducer: one Java class with the customer's pod payload embedded, run against `client-java-api` 24.0.0 / 25.0.1 / 26.0.1 / 27.0.0. Boundary comes out 24.0.0 throws, 25.0.1 throws, 26.0.1 parses, 27.0.0 parses, matching the PR exactly, and the thrown message reproduces the customer's character for character. It is attached to FD-7899 along with the full write-up.

munishchouhan:
Checked both claims independently.

**The 1.35 correction is right.** Reproduced against the upstream specs — `PodStatus.allocatedResources` is absent from release-1.33 and release-1.34 and first appears in release-1.35, together with `PodStatus.resources`:

```
release-1.32: allocatedResources=False resources=False extendedResourceClaimStatus=False
release-1.33: allocatedResources=False resources=False extendedResourceClaimStatus=False
release-1.34: allocatedResources=False resources=False extendedResourceClaimStatus=True
release-1.35: allocatedResources=True  resources=True  extendedResourceClaimStatus=True
release-1.36: allocatedResources=True  resources=True  extendedResourceClaimStatus=True
```

Fixed in a9a380e7d (the `K8sJobOperation` comment), and the PR title and description now say 1.35 with the evidence table.

One refinement on "clusters at 1.34 and below are not exposed". 1.34 is not categorically immune. It adds `PodStatus.extendedResourceClaimStatus`, and 24.0.0's `V1PodStatus` does not define that either — its full field set is:

```
conditions containerStatuses ephemeralContainerStatuses hostIP hostIPs initContainerStatuses
message nominatedNodeName observedGeneration phase podIP podIPs qosClass reason resize
resourceClaimStatuses startTime
```

So a 1.34 cluster throws the same exception for any pod that actually uses extended resource claims. Your omit-when-unset point holds and that is why BuildKit pods never hit it, so the practical exposure line is still 1.35 — but the accurate statement is "1.34 only for pods using extended resource claims", not "1.34 is safe". I have written it that way in the PR description, since it changes who needs checking.

**The 27.0.0 headroom claim also holds.** Diffed 27.0.0's `V1PodStatus` against the release-1.36 spec in both directions: zero fields missing from the model, zero extra. `nodeAllocatableResourceClaimStatuses` is present as you said.

**The `createGson()` finding is confirmed.** In the 27.0.0 jar, `io.kubernetes.client.openapi.JSON` carries 1536 `CustomTypeAdapterFactory` references and **zero** of them are inside `createGson()` — they are all registered from the static initialiser, exactly as you describe. So rebuilding the Gson from `createGson()` plus the five value adapters does drop strict validation while keeping the gson-fire preprocessor and the metadata exclusion strategy.

I am deliberately keeping that out of this PR. COMP-2368 files it as out of scope, and it is a global change to how every model in the client deserializes — trading required-field and array-shape validation across all 694 models for unknown-field tolerance. That deserves its own review and its own risk discussion rather than riding along on a version bump. Worth noting the two are complementary rather than alternatives: the lenient Gson removes the failure class, while the guard in this PR is what stops any future parse failure from being reported as a failed build. I would keep the guard even after the lenient Gson lands.

Happy to open the follow-up issue with your reproducer referenced if you have not already.

src/main/groovy/io/seqera/wave/service/job/impl/K8sJobOperation.groovy
```diff
@@ -60,18 +67,28 @@ class K8sJobOperation implements JobOperation {
         }
 
         // Find the latest created pod among the pods associated with the job
-        final pod = k8sService.getLatestPodForJob(job.operationName)
+        V1Pod pod
+        try {
```
pditommaso: The guard wraps only `getLatestPodForJob`, but `k8sService.getJobStatus(job.operationName)` earlier in this same `status()` method deserializes a `V1Job` through the identical client path and remains unguarded. If a future Kubernetes version adds an unknown field to the `V1Job`/`V1JobStatus` model — structurally the same case as the pod-level `allocatedResources` that triggered COMP-2368 — that call throws `IllegalArgumentException` outside this try/catch, escapes `status()`, and every build is again reported as failed with exit -1. Worth extending the guard to cover the Job-status parse as well, otherwise the same class of version skew reintroduces the exact bug this PR fixes.

src/main/groovy/io/seqera/wave/service/job/impl/K8sJobOperation.groovy
```diff
@@ -60,18 +67,28 @@ class K8sJobOperation implements JobOperation {
...
+            pod = k8sService.getLatestPodForJob(job.operationName)
+        }
+        catch( IllegalArgumentException e ) {
```
pditommaso: `catch(IllegalArgumentException)` is broader than the version-skew case it's meant to handle. Any `IllegalArgumentException` from `getLatestPodForJob` — a library bug, a genuinely malformed pod, an unexpected state — is converted into `jobLevelState(status, ...)` with only a `warn` log. In the `Succeeded` path that means a build reported as succeeded (exit 0) with no logs, silently masking a real failure. The comment correctly documents that `ApiException` still propagates, but consider narrowing this to the known parse-error signature (e.g. inspect the message / exception shape) or at least logging at a level loud enough that a masked failure stays diagnosable.

pditommaso (commented):
Two non-blocking observations on the pod-parse guard (COMP-2368). The fix itself is sound — catching `IllegalArgumentException` is the right type and `ApiException` correctly keeps propagating. See inline comments on the guard's scope and breadth.

src/main/groovy/io/seqera/wave/service/job/impl/K8sJobOperation.groovy
munishchouhan: Reproduced this end to end against a local k3s cluster (colima) running the real service, and the concern is concrete — but the fix it implies is not a wider try block.

**The unguarded call does throw the same way.** `getJobStatus` deserializes a `V1Job`, and a `V1Job` carries a full `V1PodTemplateSpec`, so the pod *spec* and *container* surface is parsed twice: once in the unguarded job lookup and once in the guarded pod lookup. Real 1.34 fields injected into the actual Job payload from the cluster, on `client-java-api` 24.0.0:

```
spec.template.spec.hostnameOverride                       -> IAE: The field `hostnameOverride` in the JSON string is not defined in the `V1PodSpec` properties.
spec.template.spec.containers[0].restartPolicyRules       -> IAE: The field `restartPolicyRules` in the JSON string is not defined in the `V1Container` properties.
status.<unknown field>                                    -> IAE: The field `...` in the JSON string is not defined in the `V1JobStatus` properties.
```

All three parse on 27.0.0. So a defaulted `PodSpec`/`Container` field in a future release trips `getJobStatus` before the pod lookup is reached.

**But a guard there would be a no-op.** The escape does not reach the job manager — `JobServiceImpl.status` (JobServiceImpl.groovy:136-142) already wraps `operations.status(job)` in `catch(Throwable)` and returns `JobState.unknown(t.message)`. `JobManager.state0` then holds that unknown for the grace interval and converts it to `FAILED` with a null exit. A guard around `getJobStatus` that returns unknown would produce byte-for-byte the same outcome as the catch that already exists one level up.

That is also what the pre-fix run actually did on the cluster: successful build, `Job=Complete`, image pushed, and

```
WARN  JobServiceImpl - Unable to obtain status for job=bd-7aa003800c45dc2e-1 - cause: The field `resources` ...
INFO  ContainerBuildServiceImpl - == Container build completed 'docker.io/hrma017/build:7aa...' - exit=null; status=FAILED
```

(Worth noting for the PR description too: the exception is absorbed by `JobServiceImpl`, not by the job manager as a job exception, so the reported exit is `null` via the grace-period unknown->FAILED path rather than the hard-coded `-1`.)

The fallback that would actually recover a result is the mirror of the current one: on a job-parse failure, read the carrier pod and take `containerStatuses[0].state.terminated.exitCode` plus the logs, falling back to unknown only if the pod is unparseable too. That is a change in shape rather than a wider try, which is why it reads to me as a follow-up rather than something to fold into this bump.

src/main/groovy/io/seqera/wave/service/job/impl/K8sJobOperation.groovy
munishchouhan: Checked the actual `IllegalArgumentException` surface of the client model against real payloads from a local k3s v1.35.8 cluster. The catch is broader than the skew case in a way that is measurable, and it is also narrower than skew in one direction:

```
pod status + pod-level allocatedResources (customer payload) -> IAE: The field `allocatedResources` in the JSON string is not defined in the `V1PodStatus` properties.   <- version skew
pod with containers[0].name removed                          -> IAE: The required field `name` is not found in the JSON string: {...}                                   <- NOT skew
pod with containerStatuses as an object instead of an array   -> ClassCastException: JsonObject cannot be cast to JsonArray                                             <- escapes the guard
```

So the two error classes that are not version skew are distinguishable by message shape (`is not defined in the ... properties` vs `The required field ... is not found`), which makes narrowing feasible without inspecting internals — at the cost of coupling to the generated message text.

One correction on the impact, though. In the `Succeeded` path the fallback cannot invent a success: `jobStatus0` only returns `Succeeded` when `status.succeeded` is set, which means a pod actually terminated with exit 0. What the fallback loses there is the build log, not the truth of the exit code. The `Failed` path is the lossier one — `JobState.failed(255)` discards the real exit code and the log.

For what the `Failed` path costs in practice: in one of the local runs the build genuinely failed (a 403 on the push to the configured repo) on a cluster whose pod status the client could not parse. The guard returned `JobState.failed(255)` and the buildkit output was gone, so `wave.log` carried only `exit=255; status=FAILED` - I had to go to `kubectl logs` to find the 403. That diagnosability loss is worth optimising for more than a false success is.

The logging point stands regardless: a masked non-skew parse error should be loud. The guard already increments `wave.k8s.pod.parse.errors`, and two options are open on top of that — raise the log to `error` with the exception class included so it is greppable, or additionally gate the fallback on the unknown-field message so that a malformed pod or a library bug keeps propagating instead of being absorbed. Both are small; which one goes in is your call.

munishchouhan:
Follow-up correction to my own comment above. I reproduced this end to end with the real service — local Wave (`run_local.sh`, k8s build strategy) driving builds through the wave CLI against colima clusters at both versions.

**The trigger is 1.36 by default, not 1.35.** Pod-level `status.allocatedResources`/`status.resources` exist in the release-1.35 API spec, but the kubelet only fills them in behind a feature gate — release-1.35 `pkg/kubelet/kubelet_pods.go`:

```go
if utilfeature.DefaultFeatureGate.Enabled(features.InPlacePodLevelResourcesVerticalScaling) {
    apiPodStatus.Resources = kl.convertToAPIPodLevelResourcesStatus(pod, oldPodStatus)
    apiPodStatus.AllocatedResources = resourcehelper.PodRequests(pod, opts)
}
```

and that gate in `pkg/features/kube_features.go` is `1.35: Alpha, default false` -> `1.36: Beta, default true` (its deps `InPlacePodVerticalScaling` GA 1.35, `PodLevelResources` beta 1.34, `NodeDeclaredFeatures` beta 1.36 all default-on by 1.36).

Measured, same Containerfile every time:

| cluster | pod-level fields | client 24.0.0 + pre-fix code | client 27.0.0 + this PR |
|---|---|---|---|
| k3s v1.35.8, stock | absent | build succeeds, reported succeeded | succeeded |
| k3s v1.35.8, alpha gates on | present | `Job=Complete`, image pushed, Wave reports `exit=null; status=FAILED`, CLI `succeeded:false` | `exit=0`, logs captured, CLI `succeeded:true` |
| k3s v1.36.4, **stock, no gates** | present | `Job=Complete`, image pushed 17:07:12Z, Wave reports `exit=null; status=FAILED`, CLI `succeeded:false` | `exit=0`, `{"status":"COMPLETED","succeeded":true}`, full build log |

So a stock 1.36 cluster reproduces the customer report exactly, and a stock 1.35 cluster does not reproduce it at all. The exposure line in the PR description should be 1.36+; 1.35 only with `InPlacePodLevelResourcesVerticalScaling` explicitly enabled. It follows that the customer payload proves >= 1.36, or 1.35 with that alpha gate on — which narrows the cluster question differently than we concluded above.

Two smaller notes from the same runs:

- **The field named in the exception depends on the job's resource requests.** With no `wave.build.k8s.resources.requests` configured, the build pod is BestEffort, `allocatedResources` is omitted and the validator trips on `resources` instead. Production sets requests, hence `allocatedResources` in the customer trace. Both are 1.35 pod-level additions absent from 24.0.0, so the fix is the same either way.
- **The reported exit status is `null`, not `-1`.** The exception never reaches the job manager: `JobServiceImpl.status` (:136-142) catches `Throwable` and returns `JobState.unknown`, then `JobManager.state0` converts a persistent unknown into `FAILED` with a null exit once the grace interval passes. The Cause section of the PR describes the job-manager `notifyJobException` path and a hard-coded `-1`, which is not what happens.

Nothing here changes the fix — 27.0.0 parses the 1.36 payload and the build is reported correctly on both clusters.

pditommaso (changes requested):
Thanks for the very thorough investigation and write-up — the root-cause analysis and the cluster-version table are excellent.

I'd like to propose we narrow this PR down to **only the `client-java` 24.0.0 → 27.0.0 bump** and drop the `K8sJobOperation` guard (and its tests). The bump on its own fully closes COMP-2368: 27.0.0 models the pod-level `resources`/`allocatedResources` fields, so the pod parses, the build reports `succeeded: true`, and the logs are preserved.

My concern is with the guard's failure mode rather than the code, which is clean. On a future version skew the guard turns a currently-loud failure into a **silent success with no logs**: a succeeded job would be reported `succeeded: true` with `"(logs not available…)"`, surfaced only via a WARN and the `wave.k8s.pod.parse.errors` counter. In practice that's easy to miss and harder to troubleshoot later (builds "pass" but we've silently lost build-log observability). Today, the same skew fails hard — someone notices immediately and bumps the client. I'd rather keep it loud: a hard failure on an unrecognised payload is easier to detect and diagnose than a quiet degradation.

There's also a coverage gap that makes the guard only a partial safety net: the `V1Job` parse in `getJobStatus` (which embeds a full `V1PodTemplateSpec`) isn't guarded, so a future unknown `PodSpec`/`Container` field would still throw before the pod lookup and fall back to the old failure path anyway.

So my suggestion: ship the client bump by itself now, and if we want resilience against future skew, track it separately — ideally as the permissive-deserialisation option (rebuilding the Gson from `JSON.createGson()`), which fixes the whole client uniformly rather than one call site.

Happy to discuss if I'm missing a case where the silent fallback is preferable. Thanks again!

munishchouhan:
Narrowed — the guard and its four tests are reverted, so the client bump is the whole change. Description updated to match.

Your points hold. In the local 27.0.0 runs the guard never fired (no parse warning in either log), so it contributed nothing to the `exit=0` / `succeeded:true` outcome, and the `V1Job` coverage gap is real.

One note for the follow-up, on the case you asked about. The pre-fix failure is loud in that it breaks, but it reports `succeeded:false` with a **null** exit and the exception text standing in for the build log, and because Nextflow forces `TERMINATE` on a Wave error, a skew after a cluster upgrade fails every build at once. So the trade is an outage against lost log observability — which is why the permissive-deserialisation option you suggested is the better shape than a per-call-site guard: the payload keeps parsing rather than choosing between the two, and it covers every model rather than one call site.

pditommaso (approved):

